PR #1, opened 2017-07-21
Review

josevalim:


lib/property_test.ex
```diff
@@ -5,8 +5,18 @@ defmodule PropertyTest do
...
   # TODO: moduledoc (overview of ptesting, shrinking)
 
+  # QUESTION: Should we support property and automatically include Stream.Data
+  # and check all in there? Something like:
+  #
+  #     property "foo bar and baz" do
+  #       check all ... do
+  #         ...
+  #       end
+  #     end
```
whatyouhide: Mh yeah that might be nice because importing `Stream.Data` everywhere is kind of a lot of stuff. I am 👍 on this.

lib/property_test.ex
```diff
@@ -5,8 +5,18 @@ defmodule PropertyTest do
...
+
   alias StreamData.Random
 
+  # TODO: We probably don't need this struct. A map will be fine.
```
whatyouhide: Yes this was premature architecturing 😄, let's go with a map.

lib/stream_data.ex
```diff
@@ -1,14 +1,19 @@
...
   functions exposed in this module, like `constant/1`, and by combining other
   generators through functions like `bind/2`.
 
   ## Enumerable
 
+  # TODO: The reader checking those docs for the first time I have no idea
+  # what "unshrinkable values" below means.
+  # Maybe we should write a small paragraph about when and how generators are
+  # used, i.e. generating test data and property testing. In property testing,
+  # we can talk about the importance of shrinking.
```
whatyouhide: Sounds good yes, I will add this.

lib/stream_data.ex
```diff
@@ -30,6 +35,9 @@ defmodule StreamData do
       |> Enum.take(10)
       #=> [4, 6, 4, 10, 14, 16, 4, 16, 36, 16]
 
+  # TODO: From the example above it seems generators do not emit uniq items.
+  # Maybe that should be explicitly said.
```
whatyouhide: They don't emit uniq values and yes good thing to mention it.

lib/stream_data.ex
```diff
@@ -57,6 +65,8 @@ defmodule StreamData do
   defmodule FilterTooNarrowError do
     defexception [:message]
 
+    # FIX: A good error message says what went wrong, why it went wrong
+    # and possibly what can be done to fix it.
```
whatyouhide: Whoopsie, I forgot about this. It was meant to be a `TODO: find good error message` 😄

lib/stream_data.ex
```diff
@@ -65,15 +75,26 @@ defmodule StreamData do
...
     def exception(options) do
       %__MODULE__{message: "too many duplicates: #{inspect(options)}"}
     end
   end
 
+  # TODO: All of the docs tells what happens on shrinking
+  # *before* it shows examples. I would reverse that. I am
+  # not interested on shrinking before I learn what the function
+  # does and how it works.
```
whatyouhide: Fair enough, yep.

lib/stream_data.ex
```diff
@@ -65,15 +75,26 @@ defmodule StreamData do
...
+
   ### Minimal interface
 
   ## Helpers
 
+  # QUESTION: Should we underscore those? We typically avoid
+  # such functions in STDLIB modules because import StreamData
+  # would bring them in. Alternatively we can make them public.
```
whatyouhide: Mh great point. I would start with underscoring maybe because I don't want users to know how a generator works internally for now. Wdyt?

lib/stream_data.ex
```diff
@@ -136,6 +157,11 @@ defmodule StreamData do
   end
 
   # TODO: should this be exposed?
+  # QUESTION: In Elixir, we would call this function filter_map,
+  # and have it return {:cont, value} | :skip or {:ok, value} | :error.
+  # Maybe we should standardize this return type and make this public?
+  # I would also likely call the underlying rose tree operation filter_map.
```
whatyouhide: I don't think `filter_map` is a good name here though, because we're not mapping, we're *binding*. Those are different concepts in StreamData and mapping is only a particular case of binding. But the return value, for sure I think we should standardize with elixir. Preferences?

About making it public, we can but maybe we can wait, I don't know. Preferences?

lib/stream_data.ex
```diff
@@ -296,6 +317,10 @@ defmodule StreamData do
 
   ## Generator modifiers
 
+  # TODO: We have functions about `size` but so far we haven't explained
+  # what `size` does and why it matters. Maybe we need to discuss them on
+  # the shrinking section in the module doc.
```
whatyouhide: We definitely need to discuss this, good point. Especially because contrary to what one intuitively thinks, size is not related to shrinking in any way (that is, we don't make size smaller in order to shrink). :)

lib/stream_data.ex
```diff
@@ -855,7 +883,10 @@ defmodule StreamData do
     filter(enum_data, &not(Enum.empty?(&1)))
   end
 
-  @doc """
+  # QUESTION: Should we change the arguments order of the function
+  # below? So the generator is the first arg and not a function?
```
whatyouhide: Good point. I actually made the mistake of blindly inspiring this from Clojure's implementation but they invert their arguments compared to what we do, so it totally makes sense.

lib/stream_data.ex
josevalim: Sorry, I did not meant to call it `filter_map`, I am mostly proposing to unify the function return types (and maybe call the lazy tree one filter_map).

> About making it public, we can but maybe we can wait, I don't know. Preferences?

Property calls it and it would be nice if ExUnit did not depend on internal APIs.

lib/stream_data.ex
josevalim: hahaha, see, that's why docs are important!

lib/stream_data.ex
whatyouhide: @josevalim sounds good. I like `{:cont, term} | :skip`, or `{:keep, term} | :skip` more than `{:ok, term} | :error` because there's no "error" here. Wdyt?

josevalim: @whatyouhide let's go with `{:cont, term} | :skip` then.

lib/stream_data.ex
```diff
@@ -379,14 +391,11 @@ defmodule StreamData do
     end)
   end
 
+  # QUESTION: When would I use the function below?
```
whatyouhide: It's used for things such as `byte/0` where `byte/0` depends on the implementation of `int/1` which shrinks, but it doesn't make sense for bytes to shrink. Probably, we can make this function private for now but it's worth noting that it's not possible to do what it does with the rest of the public API. Thoughts @josevalim?

josevalim: I don't think we should remove it but rather document when and how it is used.

lib/property_test.ex
```diff
@@ -1,30 +1,28 @@
...
 
-  defmodule RunOptions do
-    @moduledoc false
+  TODO: better overview of property testing and shrinking.
+  """
 
-    defstruct test_count: 100,
-              max_shrink_depth: 50
-  end
+  # TODO: how do we customize number of tests to run and such when writing
```
josevalim: Tags? We can read the tags from the config. So we can say that `property "foo bar"` defines something like:

    property "foo bar", config

And then `check all` has access to the config. The downside of this approach is that `check all` wouldn't work inside a function. So we need to go down this road carefully. But that's already the case-ish if we import Stream.Data only inside `property`. We may want to revisit that (since Stream.Data can also be used as data factory - many we need a `next` or `random` function in Stream.Data to get a random element?)

josevalim:
@whatyouhide i think this all looks great. We need to improve testing, but that can be done as part of another PR. :+1: for merging this!

Issue #7, opened 2017-07-25
Make tuples and atoms generators


whatyouhide:
It would be nice to be able to write this:

```elixir
check all result <- one_of([{:ok, binary()}, :error]), ...
```

instead of:

```elixir
check all result <- one_of([tuple({constant(:ok), binary()}), constant(:error)]), ...
```

We need to make sure that tuples and atoms are *generators*, not that they're just supported in `check/3`.

whatyouhide:
Closing in favour of the PR.

PR #10, opened 2017-07-30
Support tuples and atoms as generators

whatyouhide:
Closes https://github.com/whatyouhide/stream_data/issues/7.


lib/stream_data.ex
```diff
@@ -82,6 +82,26 @@ defmodule StreamData do
...
+  Some Elixir types are implicitly converted to `StreamData` generators when
+  used in property testing or composed. These types are:
+
+    * atoms - they generated themselves. For example, `:foo` is equivalent to
+      `StreamData.constant(:foo)`.
+
+    * tuples of generators - they generated tuples where each value is a value
+      generated by the corresponding generator, exactly like described in
+      `tuple/1`. For example, `{StreamData.int(), StreamData.boolean()}`
+      generates things like `{10, false}`.
```
josevalim: s/things/entries?

test/stream_data_test.exs
```diff
@@ -15,6 +15,20 @@ defmodule StreamDataTest do
     end)
   end
 
+  test "atoms are generators" do
+    for_many(map(:foo, &(&1)), fn term ->
+      assert term == :foo
+    end)
+  end
+
+  test "tuples are generators" do
```
josevalim: Should we add a test of a tuple with atoms inside?

josevalim:
Should we add a function that converts a data structure to a generator if possible? I know today it can be done as simple as `Stream.Data.map(arg, & &1)` though.

whatyouhide:
@josevalim fixed the two comments, thanks for pointing out. 

> Should we add a function that converts a data structure to a generator if possible? I know today it can be done as simple as Stream.Data.map(arg, & &1) though.

Yes it can and I would wait with a function that does that. Maybe we can document that you can use `map/2` to turn into a generator. Thoughts?

josevalim:
No need to document it explicitly. Let's wait until the question shows up, then we can discuss it accordingly.

PR #14, opened 2017-08-12
Use max_consecutive_failures in filter/3

manukall:


whatyouhide:
Great catch, thanks @manukall! 💟

PR #25, opened 2017-08-17
Rename int to integer

keathley:
Renames `int` to `integer`. Not sure if this is the direction that you want to go so feel free to close this PR if its not :smile:. I tried to make sure to update all the docs as well but let me know if I missed anything.


lib/stream_data.ex
```diff
@@ -931,7 +931,7 @@ defmodule StreamData do
   ## Examples
 
       data = StreamData.fixed_map(%{
-        int: StreamData.int(),
+        int: StreamData.integer(),
```
whatyouhide: Let's rename this key to `integer:` as well.

test/stream_data_test.exs
```diff
@@ -263,7 +263,7 @@ defmodule StreamDataTest do
   test "fixed_map/1" do
     data =
       fixed_map(%{
-        int: int(),
+        int: integer(),
```
whatyouhide: Let's rename this key to `integer:` as well.

keathley:
Renamed the keys. Do you need me to squash those commits?

PR #27, opened 2017-08-18
Make StreamData.list_of/1 accept length-related options

whatyouhide:


lib/stream_data.ex
```diff
@@ -701,20 +722,48 @@ defmodule StreamData do
...
       |> LazyTree.zip()
-      |> LazyTree.map(&list_lazy_tree/1)
+      |> LazyTree.map(&list_lazy_tree(&1, min_length))
       |> LazyTree.flatten()
     end)
   end
 
+  defp list_length_range_fun(options) do
+    {min, max} =
+      case options[:length] do
```
josevalim: Maybe better to use `Keyword.fetch(options, :length)`?

lib/stream_data.ex
```diff
@@ -701,20 +722,48 @@ defmodule StreamData do
   #       |> LazyTree.flatten()
   #     end)
   #
-  @spec list_of(t(a)) :: t([a]) when a: term
-  def list_of(data) do
+  @spec list_of(t(a), keyword) :: t([a]) when a: term
+  def list_of(data, options \\ []) do
+    list_length_range_fun = list_length_range_fun(Keyword.take(options, [:length, :min_length, :max_length]))
```
josevalim: No need for `Keyword.take`?

lib/stream_data.ex
```diff
@@ -701,20 +722,48 @@ defmodule StreamData do
...
+          unless (is_integer(max_length) and max_length >= 0) or max_length == :infinity do
+            raise ArgumentError, ":max_length must be a positive integer, got: #{inspect(max_length)}"
+          end
+          {min_length, max_length}
+        other ->
+          raise ArgumentError, ":length must be a positive integer or a range " <>
+                               "of positive integers, got: #{inspect(other)}"
+      end
+
+    &(min..(max |> min(_size = &1) |> max(min)))
```
josevalim: I find those pipes very unclear (because the function names and variable names). Which case is it trying to handle?

lib/stream_data.ex
whatyouhide: Ah, I did this with the idea that I would reuse this function in `uniq_list_of` which takes more options. Probably not needed in that case anyways.

lib/stream_data.ex
whatyouhide: I agree they're unclear. The cases I'm handling are:

* if the max length is > than the size, we still want to have the size limit the length so that we still have "growing" lists instead of having possibly big lists with a small size
* if however the size is < than the min length, we want to generate a list of exactly the min length because if we listened to the size, we would break the min lenght

Is it clear now? If so, any ideas on how to improve it?

lib/stream_data.ex
whatyouhide: Not a big difference to me but why not :)

lib/stream_data.ex
josevalim: Maybe `fn size -> min..(max |> min(size) |> max(min) end`?

whatyouhide: Oh sure, I guess capture syntax didn't help here 😄

lib/stream_data.ex
josevalim: The fetch variant does not allow `length: nil` ;)

whatyouhide: Yeah yeah, problem is that we still allow it for `:min_length` and `:max_length` :|

whatyouhide:
@josevalim updated with a bunch of changes, only question left then is if `:max_lenght` and `:min_length` are fine as `nil`.

josevalim:
@whatyouhide we can use `Keyword.get(options, :max_length, :infinity)` then. :)

PR #31, opened 2017-08-25
Fix typos in the PropertyTest @moduledoc

jeffkreeftmeijer:
I found some typos while reading the [PropertyTest](https://hexdocs.pm/stream_data/PropertyTest.html) documentation, so I'd like to break the ice with the obligatory typo patch. Thanks for your work on StreamData, @whatyouhide, et al.!


whatyouhide:
Thanks @jeffkreeftmeijer! 💟

PR #41, opened 2017-09-18
Introduce StreamData.optional_map/1

whatyouhide:
Had some downtime during lunch and gave this a try.

@josevalim the reason I went with a manual implementation is to make the following things happen:

1. I wanted to generate the smallest possible amount of values for the map, so that if I have `optional_map(%{some_key: very_expensive_generator})` then `very_expensive_generator` only generates when strictly needed
1. I wanted that given a generated map, it shrinks by shrinking the values and then taking out keys but without generating new smaller maps when taking out elements, only shrinking existing elements. This means that with this implementation, if `optional_map(%{a: boolean(), b: integer()})` generates `%{a: false, b: 0}`, we don't shrink to `%{a: true}` ever (which could happen if we generate a new map when we take out keys).

All the stuff mentioned in point 2. makes sense if you look at how simple would be an implementation that generates new maps when taking out keys instead:

```elixir
def optional_map(data_map) do
  bind(sublist(Map.keys(data_map)), fn keys ->
    fixed_map(Map.take(data_map, keys))
  end)
end
```

With the implementation above, however, when we shrink the keys resulted by `sublist`, we need to generate a new map which will then have completely new values to shrink. I think a better design decision is to behave like I implemented currently and not like shown above. Maybe there's a way to implement what I did in the PR in a much simpler way, but I couldn't find any, so help is very welcome :)

Thoughts?


lib/stream_data.ex
```diff
@@ -1032,6 +1032,53 @@ defmodule StreamData do
...
+
+      call(fixed_map(base_data_map), seed2, size)
+      |> LazyTree.map(fn fixed_map -> LazyTree.map(subkeys_tree, &Map.take(fixed_map, &1)) end)
+      |> LazyTree.flatten()
+    end)
+  end
+
+  defp sublist(list) do
+    map(list_of(boolean(), length: length(list)), fn indexes_to_keep ->
+      for {elem, keep?} when keep? <- Enum.zip(list, indexes_to_keep), do: elem
```
michalmuskala: `{elem, true} <- ...` would work the same

michalmuskala:
@whatyouhide I wonder if maybe something like this would work:

```elixir
def optional_map(data_map) when is_map(data_map) do
  data_map
  |> Map.keys
  |> sublist()
  |> map(fn keys ->
    fixed_map(Map.take(data_map, keys))
  end)
end
```

This has the advantage of never generating values it doesn't need, but I might be missing something.

whatyouhide:
@michalmuskala first of all, the fun passed to `map/2` has to return a value not a generator. The one that returns a generator is `bind/2`, and if you take your implementation and replace `map/2` with `bind/2`, you get exactly the implementation that I posted in the issue, just with less pipes:

```elixir
def optional_map(data_map) do
  bind(sublist(Map.keys(data_map)), fn keys ->
    fixed_map(Map.take(data_map, keys))
  end)
end
```

The problem with this implementation is: say we have `optional_map(%{a: boolean(), b: integer()})`. We generate `%{a: false, b: 1}` and tests fail, so we start shrinking. The first shrink happens by shrinking the result of `fixed_map/1` (that's how `bind/2` works, it first shrinks the returned generator and then the original generator). When shrinking the result of `fixed_map/1`, we might go to `%{a: false, b: 0}` and that's it because `integer/0` shrinks towards `0` and `boolean/0` towards `false`. So now we shrink the generator passed as the first argument to `bind/2`, which is `sublist/1` which shrinks towards `[]`. It had generated `[:a, :b]` so say it shrinks as `[:a]`. Now we need to recreate a completely different `fixed_map(Map.take(data_map, :a)) == fixed_map(%{a: boolean()})`. We now can generated `%{a: true}`, but this is not a good shrink because we "unshrinked" the value we wanted to shrink. Eventually you'll have `%{a: false}` (which is what we want to have by just taking keys out) as well but it's very suboptimal (and `boolean/0`) is not a particularly good example because it has a space of only two values. Does this clarify?

michalmuskala:
Yes, thank you for the explanation. 

Viewing this from another angle, if I understand you correctly means the `optional_map` generator as it is implemented right now would never produce the `%{a: true}` value - is that right?

whatyouhide:
@michalmuskala exactly right! It would only *shrink* values from the map it generates, as well as taking out key-value pairs from that same map. So if you had %{a: false, b: 1}`, it would shrink to `%{a: false, b: 0}`, `%{a: false}`, `%{b: 1}`, `%{b: 0}`, and `%{}`, without ever shrinking to say `%{a: true}` or `%{b: 2}` or anything like that.

michalmuskala:
Then, to be honest, I'm not sure this is a good implementation. There might be cases where values bigger than minimal trigger bugs. A simple case where the current implementation of `optional_map` would never hit the bug would be:

```elixir
def foo(%{a: _, b: _}), do: :ok
def foo(%{a: 42}), do: raise "that's a wrong answer!"
```

whatyouhide:
@michalmuskala I am not sure I follow 😕 For the `foo/1` function above, this implementation of `optional_map/1` will eventually generate either `%{a: 42}` from scratch, or `%{a: n} when n >= 42` from scratch which will eventually shrink to `%{a: 42}`, or `%{a: n, b: anything} when n >= 42` which will shrink to a map that matches `%{a: 42}`. See what I mean?

michalmuskala:
Ah, I think I misunderstood how the shrinking algorithm works - I forgot it will backtrack when it shrinks to the minimum and doesn't hit the bug.

PR #43, opened 2017-09-22
Allow :rand.export_seed/1 format in check_all

josevalim:


test/stream_data_test.exs
```diff
@@ -512,6 +512,21 @@ defmodule StreamDataTest do
     assert check_all(list_of(boolean()), options, property) == {:ok, %{}}
   end
 
+  test "check_all/3 with :rand.export_seed()" do
+    :rand.uniform()
```
whatyouhide: @josevalim is this to set a seed? If so, what would be the downside of being explicit and doing:

```elixir
seed = :rand.seed_s(:exs1024)
options = [initial_seed: :rand.export_seed_s(seed)
```

?

josevalim: No downside, that's better, yes!

josevalim:
@whatyouhide btw, please take a look at the build, it has some races!

PR #50, opened 2017-10-15
Fix the tree docs

alfert:
Updated the docs for `tree`: 
* differentiate between generating leafs and inner nodes
* additional example 
* spec uses parameter names to be more clear


lib/stream_data.ex
```diff
@@ -1276,31 +1272,6 @@ defmodule StreamData do
       Enum.at(StreamData.resize(data, 10), 0)
       #=> [[], '\t', '\a', [1, 2], -3, [-7, [10]]]
 
-  Another common example is a tree, where the data exists only in the inner nodes.
```
alfert: Apparently you dislike my example :-)

lib/stream_data.ex
```diff
@@ -1250,21 +1250,17 @@ defmodule StreamData do
         defstruct [:left, :right]
       end
 
-  We can start off by creating a generator that generates branches given the
-  generator that generates the content of each node (`integer/0` in our case):
+  Now, we can generate trees by using the `integer()` generator to generate
```
alfert: This is way better than before! 👍

lib/stream_data.ex
whatyouhide: It's just the same example as above so repeating two trivial "tree" examples seemed very redundant :) I tried taking the improvements to the documentation you made and use them in the descriptive parts.

whatyouhide:
@alfert we use "Examples" like we use "Options" even when we have just one example/option. That's the title of the section "Examples" which is a section containing the examples (which sometimes are incidentally just one).

alfert:
:+1:

PR #52, opened 2017-10-16
Add ExUnitProperties.pick/1

whatyouhide:


lib/ex_unit_properties.ex
```diff
@@ -471,4 +471,28 @@ defmodule ExUnitProperties do
...
+
+  ## Examples
+
+      ExUnitProperties.pick(StreamData.integer())
+      #=> -21
+
+  """
+  @spec pick(StreamData.t(a)) :: a when a: term()
+  def pick(data) do
+    seed = ExUnit.configuration()[:seed]
```
josevalim: Let's use the new seed in ExUnit v1.6?

whatyouhide: You mean the seed in the pdict? I guess that would make more sense?

josevalim:
Yeah. The question is what to do when they call it outside of the test
process. Raise, right?
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

josevalim:
We cannot use the global seed because then all tests will likely pick the
same value :P
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

whatyouhide:
@josevalim updated :)

PR #54, opened 2017-10-18
Fix stacktraces for filters in "gen all"/"check all"

### LINKED ISSUE #46: Investigate proper stacktraces for filters in check all / gen all

whatyouhide:
Closes #46.

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -227,12 +227,12 @@ defmodule ExUnitProperties do
   defp compile(clauses, body) do
     quote do
       var!(generated_values, unquote(__MODULE__)) = []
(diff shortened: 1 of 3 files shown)


lib/ex_unit_properties.ex
```diff
@@ -241,33 +241,52 @@ defmodule ExUnitProperties do
...
               [
                 {unquote(Macro.to_string(clause)), generated_value}
                 | var!(generated_values, unquote(__MODULE__))
               ]
 
-            unquote(compile_clauses(rest, body))
-          _other ->
+            unquote(compile_clauses(rest, body, line))
+
+          other, tries_left = 1 ->
```
josevalim: Unused variable (albeit we don't warn for generated code).

PR #55, opened 2017-10-24
fail_eventually

alfert:
The `fail_eventually` macro allows to state properties that are expected to eventually fail.

The `check all` macro is essentially an universal quantifier (for all x in X the following holds..), `fail_eventually` is its existential counterpart: there is some x in X for that following does not hold. It wraps a `check all` property and test whether that property fails. If yes, then everything is ok, otherwise it fails. 

Technically, the same can be achieved by wrapping the `check all` macro with an `assert_raise`, as already done in the tests. But this depends on the internals of `check all` and we need to catch two exceptions: `ExUnitProperties.Error` and `ExUnit.AssertionError`.  With `fail_eventually` the programmer can express her intention more clearly and there is no need to negate the property.

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -106,7 +106,8 @@ defmodule ExUnitProperties do
   @doc false
   defmacro __using__(_opts) do
     quote do
-      import unquote(__MODULE__), only: [property: 2, property: 3, check: 2, gen: 2]
+      import unquote(__MODULE__), only: [property: 2, property: 3, check: 2, gen: 2,
+          fail_eventually: 1]
       import StreamData
     end
   end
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @alfert! Here are some other options to implement this:

1. Add a `fail_eventually: true` option to `check_all`
2. Have a `fail_eventually` function instead of macro where the first argument is an anonymous function, similar to `assert_raise`

I am not saying any of those is preferable yet but I thought I would contribute to the discussion. :)

alfert:
@josevalim I prefer functions generally, so the 2nd alternative is a nice option. However, my impression from the discussion about the aggregation functions was that an approach which is not based on function composition is preferred. This lead me into this direction. The functional implementation is equally simple.

Your first approach is an interesting option. It allows to reduce the level of nesting, which is here not a benefit. It would look like this, I assume: 

```elixir
check all 
     n <- integer(), 
     fail_eventually: true do
   assert n >= 0
end
```

josevalim:
Another question is: do other libraries provide similar? If so, what is the
name and its args?
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

whatyouhide:
I'd like to take a step back and ask the "deeper" meaning of this functionality.

By reading the documentation you wrote and the examples, it looks like this function should be used to test properties of the generators rather than properties of the tested code. For example, in the `integer()`/`assert n >= 0` example, you're basically testing that some integers generated by the `integer()` generator are positive while others are negative. The same goes for the `hd/1` example: you're testing that `list_of/1` generates empty and non-empty lists? I might be thinking like this because the examples are trivial, in that case sorry :P I always thought of property-based testing as testing that a property holds (which is not true in your example because the property should not hold in all cases) for the **right data**. With `fail_eventually` it seems we're saying "we'll give you only some right data and some wrong data, and you should fail sometimes" (we don't even specify that you should fail on *wrong* data, only that you should fail sometimes). It's a bit confusing to me, can you help clarify? :)

alfert:
@josevalim, in PropEr and thus prop_check there is the `fails` wrapper, which wraps a property. `fail_eventually` is clone of its semantics. 

In Erlang QuickCheck exists the macro `sometimes`, which is different and is used in the context of Pulse, i.e. for detecting race conditions: `sometimes` allows that the property may fail due to non-deterministic behaviour. In that case, the same test is repeated. The additional parameter, `n`, stating how often the property is tested again and again. If it fails every time, then the entire property fails.

@whatyouhide, your are right, my examples are really trivial. But I wanted to be sure that all feature work and are predictable, hence the trivial examples. But you are right, that my docs don't answer the question for the use case. I used it for having a successful property for a planned buggy function implementation to show how and thought it would be helpful. In the last hours, I researched the examples and tutorials of PropEr, but did not find any use of `fails`. In my memory, it was used for stateful testing, but this seems  not to be the case. 

While the addition of `fail_eventually` has some formal beauty as it is the negation of `check all`, but to have a test that shows that a property fails can be made simpler by an `assert_raise`. To allow sloppy properties because you use `fail_eventually` is not a good advice and we should not foster it by providing it in the API. 

Guys, thank you for your comments, they gave me some food for thought and let me research deeper the existing systems. 

I will close this PR.

PR #56, opened 2017-10-26
Fix typo into docs

xadhoom:


whatyouhide:
Great catch @xadhoom, thanks! 💟

PR #68, opened 2017-12-10
Improve speed of the integer/0,1 generators

whatyouhide:


lib/stream_data.ex
```diff
@@ -505,18 +505,31 @@ defmodule StreamData do
...
+
+  defp integer_lazy_tree(int, range, current, {:cont, acc}, fun) do
+    to_emit = int - current
+
+    cond do
+      to_emit == int ->
+        {:done, acc}
 
-    lazy_tree(int, children)
+      to_emit in range ->
```
michalmuskala: If we're going for max speed here, it might make sense to deconstruct the rage and pass `start, end` in the loop - this should save a protocol dispatch.

whatyouhide: @michalmuskala this PR is directed at speeding up the creation of the lazy tree returned by `integer/0,1`, not the "actuation" of said lazy tree. We only "realize" this tree when we shrink and go down the tree, otherwise we only use the root; however, creating a few streams turns out to be way more expensive than returning a function (`&integer_lazy_tree(...)`) that does the streaming. Hope I made this a bit clearer :)

PR #70, opened 2017-12-16
Improve docs

michalmuskala:
* make `ExUnitProperties.__using__` not private - otherwise there's no way
to know if it use should or should not be called by users
* make the power_of_two_with_zero function private


whatyouhide:
Thanks @michalmuskala! 💟

PR #72, opened 2017-12-18
Add `max_run_time` configuration option

### LINKED ISSUE #69: Allow for max_run_time as well as max_runs
We were discussing property testing a bit last night at our local Elixir meetup here in Berlin, and someone brought up the idea of allowing for a `max_run_time` option as well as `max_runs` to help give a more human-understandable expectation for how long a given property test might run. We thought this might be especially helpful in CI.

As to an implementation, I figure it wouldn't be super hard since this is a very similar feature compared to how we know how long to run benchmarks in Benchee. I'd be happy to help with adding this if you think this is a nice feature to add.

devonestes:
This allows users to set both a `max_runs` option to their property
tests, limiting the tests to a set number of generations, as well as a
`max_run_time` option to ensure that the tests don't take longer than a
certain time period. This is the first rough draft of this feature, and
I'm going to focus more on the documentation and some refactoring in
the next commit after I get a little feedback from folks. As it is now
it's functional, though.

Resolves #69

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -181,6 +181,14 @@ defmodule StreamData do
     end
   end
 
+  defmodule InfiniteLoopError do
+    defexception message: """
+    You have attempted to set the `max_runs` and `max_run_time` settings both to
+    `:infinity`, which will result in an infinte loop. Please be sure to set at least
+    one of these settings to avoid this error.
+    """
+  end
+
   ### Minimal interface
 
   ## Helpers
@@ -1935,10 +1943,13 @@ defmodule StreamData do
 
     config =
       case {Keyword.get(options, :max_runs), Keyword.get(options, :max_run_time, :infinity)} do
+        {:infinity, :infinity} ->
+          raise StreamData.InfiniteLoopError
+
         {max_runs, :infinity} ->
           Map.put(config, :max_runs, max_runs || 100)
 
-        {nil, max_run_time} ->
+        {:infinity, max_run_time} ->
           Map.put(config, :max_run_time, start_time + max_run_time)
 
         {max_runs, max_run_time} ->
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -1930,11 +1930,22 @@ defmodule StreamData do
     seed = new_seed(Keyword.fetch!(options, :initial_seed))
     size = Keyword.get(options, :initial_size, 1)
     max_shrinking_steps = Keyword.get(options, :max_shrinking_steps, 100)
-    max_runs = Keyword.get(options, :max_runs, :not_set)
-    max_run_time = Keyword.get(options, :max_run_time, :infinity)
-    config = generate_config(max_runs, max_shrinking_steps, max_run_time)
+    start_time = System.system_time(:millisecond)
+    config = %{max_shrinking_steps: max_shrinking_steps}
 
-    check_all(data, seed, size, fun, _runs = 0, System.system_time(:millisecond), config)
+    config =
+      case {Keyword.get(options, :max_runs), Keyword.get(options, :max_run_time, :infinity)} do
+        {max_runs, :infinity} ->
+          Map.put(config, :max_runs, max_runs || 100)
+
+        {nil, max_run_time} ->
+          Map.put(config, :max_run_time, start_time + max_run_time)
+
+        {max_runs, max_run_time} ->
+          Map.merge(config, %{max_run_time: start_time + max_run_time, max_runs: max_runs})
+      end
+
+    check_all(data, seed, size, fun, _runs = 0, start_time, config)
   end
 
   defp check_all(_data, _seed, _size, _fun, _runs, current_time, %{max_run_time: end_time}) when current_time >= end_time do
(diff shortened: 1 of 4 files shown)


lib/stream_data.ex
```diff
@@ -1956,6 +1963,39 @@ defmodule StreamData do
     end
   end
 
+  # max_runs and max_run_time both not set by user, default to 100 runs.
```
devonestes: I left all my comments in (that were there to help me plan stuff out) because I figured it would make it easier to review. I plan on taking them out in the next commit.

lib/stream_data.ex
```diff
@@ -1924,27 +1930,28 @@ defmodule StreamData do
     seed = new_seed(Keyword.fetch!(options, :initial_seed))
     size = Keyword.get(options, :initial_size, 1)
     max_shrinking_steps = Keyword.get(options, :max_shrinking_steps, 100)
-    max_runs = Keyword.get(options, :max_runs, 100)
+    max_runs = Keyword.get(options, :max_runs, :not_set)
+    max_run_time = Keyword.get(options, :max_run_time, :infinity)
+    config = generate_config(max_runs, max_shrinking_steps, max_run_time)
```
whatyouhide: I would do something like:

```elixir
start_time = System.system_time(:millisecond)
config = %{max_shrinking_steps: max_shrinking_steps}

config =
  case {Keyword.get(options, :max_runs), Keyword.get(options, :max_run_time, :infinity)} do
    {max_runs, :infinity} ->
      Map.put(config, :max_runs, max_runs || 100)
    {nil, max_run_time} ->
      Map.put(config, :max_end_time, start_time + max_run_time)
    {max_runs, max_run_time} ->
      Map.merge(config, %{max_end_time: start_time + max_run_time, max_runs: max_runs}) 
  end

check_all(data, seed, size, fun, _runs = 0, start_time, config)
```

To compress code a little bit. Wdyt?

test/ex_unit_properties_test.exs
```diff
@@ -72,6 +74,52 @@ defmodule ExUnitPropertiesTest do
       assert Agent.get(counter, & &1) == 10
     end
 
+    property "runs for the specified number of milliseconds" do
+      {:ok, counter} = Agent.start_link(fn -> 0 end)
+
+      # I'm explicitly passing max_runs: :not_set to override the project-wide
+      # config in the `mix.exs` file of `max_runs: 100`.
+      check all :ok <- :ok, max_runs: :not_set, max_run_time: 250 do
```
whatyouhide: I think `nil` is a better default for `max_runs`, as I used in the snippet of code above.

test/ex_unit_properties_test.exs
```diff
@@ -72,6 +74,52 @@ defmodule ExUnitPropertiesTest do
...
+        :timer.sleep(1)
+        Agent.update(counter, &(&1 + 1))
```
whatyouhide: What if you use `max_run_time: 250` and use `Process.sleep(100)` inside the block, and then assert that we did exactly two runs? It should be much more precise when using bigger intervals than something like 1ms.

test/ex_unit_properties_test.exs
devonestes: I started with `nil` as the default, but the problem I ran into there was this line: https://github.com/whatyouhide/stream_data/pull/72/files#diff-be707432787f16f5351a8cc945bdb9cbR387

When we passed in `nil` as an option it would be falsey, and so we default to the application environment setting, which meant I couldn't expressly override the application configuration for this test.

We could change that line to use `Keyword.has_key?/2` to set the default so that we can override with `nil` by passing it in if you think that's preferable.

lib/stream_data.ex
devonestes: Yeah, works for me. I originally had something like this, but it seemed like a big enough piece of logic to warrant its own function, which is why I extracted it. I'm happy with this if you'd prefer.

test/ex_unit_properties_test.exs
devonestes: I wanted a case with more than 100 runs to make sure that we're not accidentally cutting off at 100 runs even if we shouldn't be (since I almost accidentally did that while implementing this feature). It just made me feel safe, but maybe it's overkill? If you think it's overkill then I think we can get rid of this entire test and just rely on the other test since that covers things pretty well.

whatyouhide:
I was thinking something else. You're right about the `nil` handling shit. I think the way we can solve that is to simply not set `max_runs: 100` in the default app env (`mix.exs`). In there, we can instead have something like `default_max_runs: 100` or even have this hardcoded as `@default_max_runs` in `stream_data.ex`. Then, we can do something like:

```elixir
max_runs = Keyword.get_lazy(options, :max_runs, fn -> Application.get_env(:stream_data, :max_runs) end)
```

and if it's still not set (that is, the user didn't provide `:max_runs` in neither the `options` nor the app's env, we can go with the actual default (`@default_max_runs` or the app's env's `default_max_runs`). Wdyt?

devonestes:
Yeah, I think `max_runs: :infinity` is kind of what I was going for with `max_runs: :not_set`. That way you can have some sort of global configuration that applies to 99% of your cases, but you still have the power to override it in specific examples if you want. `:infinity` might be more clear as to what would actually happen, though. Plus the best part is technically that's already implemented!

Because it helps me to think about these types of things, I made a little chart with the possible settings and the behavior that we might expect given those settings. Think this is along the lines of the behavior you want?

| application env | :max_runs | :max_run_time  | behavior |
| ------------- |-------------| -----| ------ |
| not set | not set | not set | 100 runs |
| not set | :infinity | not set |  Infinite loop! 🔁  |
| max_runs: 200, max_run_time: 2000 | not set | not set | 200 runs or 2000 milliseconds (whichever happens first) |
|  max_runs: 200, max_run_time: 2000 | :infinity | not set | 2000 milliseconds |
|  max_runs: 200, max_run_time: 2000 | not set | :infinity | 200 runs |
|  max_runs: 200, max_run_time: 2000 | :infinity | :infinity | Infinite loop! 🔁  |

My bet is we probably want to avoid those infinite loop possibilities, right?

whatyouhide:
Hey @devonestes, sorry for the late reply!

> My bet is we probably want to avoid those infinite loop possibilities, right?

Yeah, those should likely be errors. One more thing we can do is default the `max_run_time` in the application's environment to `:infinity`.

whatyouhide:
Hey @devonestes, sorry for the late reply. I made some minor changes myself on top of your code. This looks nice! Thanks for your contribution :) 💟

Issue #51, opened 2017-10-15
Statefull generator combinators


alfert:
I would like to see more combinators to create more complex data, in particular a reducer to create data depending on the already generated elements.

Background:

The reducers could be a nice solution for generating command sequences to support symbolic state machines for property testing of stateful systems. Here we start with an initial state and generate commands depending on the current state and producing a new state. A reduce or unfold function could be a solution for that problem.

If this goes along your development direction of StreamData, I can work on PRs for that.

josevalim:
@alfert this sounds very exciting but I am having some difficulty in seeing some of the use cases. Could you please expand on that area? Thank you!

alfert:
@josevalim sure: property testing of stateful systems in Quickcheck, Proper, and others work with a state machine as the abstraction of the system under test (SUT) and a command generator, which generates a list of commands against the SUT. Pre- and postconditions on the state machine define valid command sequences, state transition of the state machine and connect SUT values with their state machine counter parts. There are two phases in stateful system testing: the first phase generate valid command sequences depending on the preconditions of the state machine, the second phase uses the generated command sequences and executes it against the SUT checking the postcondition of the state machine is the same as the results of the SUT. A failing property results in a shrinked command list to get the minimal sequence of commands that exhibit the error. 

Use cases for this approach are property based testing of all kinds of stateful (sub) systems, such as GenServers, processes, but also data structures (see "breaking erlang maps" below). Another interesting approach is to run a long sequence of commands against a system, e.g. a sequence of 3.000 or more commands. These are tests which you would never write as a unit test and the property testing approach guarantees that each command execution sequence is different. 

Some documentation and examples are here:
* https://medium.com/@jlouis666/breaking-erlang-maps-2-362730a91400
* http://propertesting.com/book_stateful_properties.html
* http://proper.softlab.ntua.gr/Tutorials/PropEr_testing_of_generic_servers.html
* http://www.quviq.com/getting-better-statistics-from-state-machine-tests/
* https://fscheck.github.io/FsCheck/StatefulTesting.html

josevalim:
Thanks @alfert. I thought you were hinting at something else when you said about relying on reduce. :) Do you have a proposal about the API? Maybe something you already leverage in propcheck?

alfert:
@josevalim Depends on what kind of API you are thinking about. 

[...]

PR #76, opened 2017-12-29
Unfold Combinator

alfert:
The `unfold` combinator is similar to `Stream.unfold/2` and solves #51. The combinator generates lists using an accumulator for generating the list's elements. This provides the basis for command generation used in state-machine based property testing of stateful systems, requiring to track a symbolic state during command generation.


lib/stream_data.ex
```diff
@@ -823,37 +823,10 @@ defmodule StreamData do
...
-
-      data
-      |> call_n_times(next_seed, size, length, [])
-      |> LazyTree.zip()
-      |> LazyTree.map(&list_lazy_tree(&1, 0))
-      |> LazyTree.flatten()
-    end)
+  @spec list_of(t(a), keyword()) :: t([a]) when a: term()
+  def list_of(data, options \\ []) do
+    unfold(nil, fn _ -> {data, nil} end, options)
```
michalmuskala: Please verify how this compares performance-wise to the previous implementation. `list_of` is used in many other combinators, so it heavily impacts overall performance of the library.

alfert: Good point. Do you have a kind of testbed and performance data as bench mark from your performance optimisation activities? Would be nice to do it the same way.

alfert: First results running 
`MIX_ENV=prod mix profile.fprof -e 'StreamData.list_of(StreamData.integer()) |> Enum.take(1000)'`

Test | Function | ACC(ms) | OWN(ms) | Percentage(acc)
-----|---------|-----------|-----------|---
`list_of` | `:fprof.apply_start_stop/4` | 1811.293  | 0.008 | 100.00 %
`list_of`  | anonymous `fn/3` in `StreamData.list_of/1` | 1783.694 | 8.326 | 100.00 %
`unfold`| `:fprof.apply_start_stop/4` | 2102.877 | 0.010 | 116.09 %
`unfold`|  anonymous `fn/4` in `StreamData.unfold/2` |  2073.250 | 15.602 | 116.23 %

The additional indirection of `unfold` seems to cost 16% more performance. Looks like I eat your performance optimisations. So `list_of` should not be implemented in terms of `unfold`. I will correct that.

alfert:
I close this pull request, since it was never merged nor did the shrinking job properly.

PR #77, opened 2018-01-20
Document options for `string/2`

axelson:
Also fix incorrect references to non-existant functions `list_of/3` and `uniq_list_of/3`.

**Question:**
Is "See the documentation of `list_of/2` for the possible values of options." better than duplicating the list of options again? Should the documention of other places that `list_of/2` is called be changed as well?


whatyouhide:
Thanks @axelson! 💟

axelson:
Great, thanks!

PR #79, opened 2018-02-01
Remove implementation details from StreamData docs

josevalim:
The initial description talks about wrapper values that are never
really exposed to the developer, so we simply don't care about
them. 

This PR removes those cases and aims to simplify the `@moduledoc`
of the `StreamData` module.


josevalim:
@whatyouhide free to do any change before/during/after merging. :)

whatyouhide:
Great job @josevalim, thanks! 💟

PR #82, opened 2018-02-03
Add support for `do: body` syntax.

NJichev:
Fixed a small TODO :)
Checking the AST it was of the form:

 ```
 {:gen, _meta,
   [
      {:all, _meta,
        [
          ...clauses...,
          [do: body]
        ]
      }
   ]
 }
```

Decided to just reverse it compile time and get the body.

Wasn't sure how to test it, so added a one clause test that it compiles
and one that the generators are fine since I do a reverse there.

Tell me if I've missed anything, or perhaps should change the docs.

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -223,7 +223,8 @@ defmodule ExUnitProperties do
 
   """
   defmacro gen({:all, _meta, clauses_with_body}) do
-    [[do: body] | clauses] = Enum.reverse(clauses_with_body)
+    [[do: body] | reversed_clauses] = Enum.reverse(clauses_with_body)
+    clauses = Enum.reverse(reversed_clauses)
     compile(clauses, body)
   end
 
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -374,8 +374,18 @@ defmodule ExUnitProperties do
       end
 
   """
+  defmacro check({:all, _meta, clauses_with_body}) when is_list(clauses_with_body) do
+    [[do: body] | options_and_clauses] = Enum.reverse(clauses_with_body)
+    clauses_and_options = Enum.reverse(options_and_clauses)
+    compile_check_all(clauses_and_options, body)
+  end
+
   defmacro check({:all, _meta, clauses_and_options} = _generation_clauses, [do: body] = _block)
            when is_list(clauses_and_options) do
+    compile_check_all(clauses_and_options, body)
+  end
+
+  defp compile_check_all(clauses_and_options, body) do
     {clauses, options} = split_clauses_and_options(clauses_and_options)
 
     quote do
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -222,7 +222,11 @@ defmodule ExUnitProperties do
     * value generation clauses affect shrinking similarly to `bind/2`
 
   """
-  # TODO: support do: syntax
+  defmacro gen({:all, _meta, clauses_with_body}) do
+    [[do: body] | clauses] = Enum.reverse(clauses_with_body)
+    compile(clauses, body)
+  end
+
   defmacro gen({:all, _meta, clauses} = _generation_clauses, [do: body] = _block) do
     compile(clauses, body)
   end
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -2076,8 +2076,10 @@ defmodule StreamData do
     :rand.seed_s(exported_seed)
   end
 
-  @compile {:inline,
-            split_seed: 1, order: 2, uniform_in_range: 3, lazy_tree: 2, lazy_tree_constant: 1}
+  @compile {
+             :inline,
+             split_seed: 1, order: 2, uniform_in_range: 3, lazy_tree: 2, lazy_tree_constant: 1
+           }
 
   defp split_seed(seed) do
     {int, seed} = :rand.uniform_s(1_000_000_000, seed)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by whatyouhide** [comment]
We would need to support this in `check all` as well. Can you implement that? :)

**Comment by NJichev** [comment]
A small note, when I was testing the `do:` thingy with this example from the docs:
```
      check all int1 <- integer(),
                int2 <- integer(),
                int1 > 0 and int2 > 0,
                sum = int1 + int2 do
        assert sum > int1
        assert sum > int2
      end
```
was getting too many times FilterTooNarrow, this is on:
```
λ  elixir -v                                                                         stream_data  add-do-syntax-to-gen-all
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]
Elixir 1.7.0-dev (e4a82d5f8) (compiled with OTP 20)
```
Also dunno why mix format fails, locally it passes :< (maybe elixir version)

File: lib/ex_unit_properties.ex [hunk 1]
```diff
@@ -222,7 +222,11 @@ defmodule ExUnitProperties do
     * value generation clauses affect shrinking similarly to `bind/2`
 
   """
-  # TODO: support do: syntax
+  defmacro gen({:all, _meta, clauses_with_body}) do
+    [[do: body] | clauses] = Enum.reverse(clauses_with_body)
```
whatyouhide: Don't we need to `Enum.reverse(clauses)` to bring the clauses back to the original order here as well?

NJichev: My understanding is, that there are no options in the `gen all` macro and there are only generation clauses for types of data. And the order you give the clauses is the reverse of the order the variables are injected through `var!` into the body's binding. Reversing them here would only cause difference if for some reason you have this:
```
      data = gen all int <- binary(),
                             int <- integer(), do: int

      check all int <- data do
        assert is_integer(int)
      end
```
So this would fail, since we have reversed the order of the clauses and the second integer is actually injected first and being overwritten by the first one. I can do a second reverse, so that there is minimum surprise there?

whatyouhide: Subsequent clauses can depend on previous clauses:

```elixir
data = gen all list <- list_of(integer()),
               member <- member_of(list),
               do: {list, member}
```

This would not work without reversing the clauses back to the original order.

NJichev: Oh, right :sweat_smile: , anyway, added it, should I add some tests for that too?

whatyouhide:
Thanks @NJichev! 💟

lib/ex_unit_properties.ex
```diff
@@ -222,7 +222,12 @@ defmodule ExUnitProperties do
     * value generation clauses affect shrinking similarly to `bind/2`
 
   """
-  # TODO: support do: syntax
+  defmacro gen({:all, _meta, clauses_with_body}) do
+    [[do: body] | reversed_clauses] = Enum.reverse(clauses_with_body)
```
josevalim: `{clauses, [[do: body]]} = Enum.drop(clauses_with_body, -1)` would do it too.

lib/ex_unit_properties.ex
```diff
@@ -370,8 +375,18 @@ defmodule ExUnitProperties do
       end
 
   """
+  defmacro check({:all, _meta, clauses_with_body}) when is_list(clauses_with_body) do
+    [[do: body] | options_and_clauses] = Enum.reverse(clauses_with_body)
```
josevalim: Same here.

lib/stream_data.ex
```diff
@@ -2076,8 +2076,10 @@ defmodule StreamData do
     :rand.seed_s(exported_seed)
   end
 
-  @compile {:inline,
-            split_seed: 1, order: 2, uniform_in_range: 3, lazy_tree: 2, lazy_tree_constant: 1}
+  @compile {
+             :inline,
```
josevalim: Is this being caused by different Elixir versions?

whatyouhide: Yes, I'm fixing that.

lib/ex_unit_properties.ex
whatyouhide: Good catch!

test/ex_unit_properties_test.exs
```diff
@@ -33,6 +33,22 @@ defmodule ExUnitPropertiesTest do
         Enum.take(data, 1)
       end
     end
+
+    test "supports do keyword syntax" do
+      gen(all _boolean <- boolean(), do: :ok)
```
josevalim: We should add `gen` to a local `.formatter.exs` (or to Elixir one) so we can get rid of parens.

lib/ex_unit_properties.ex
NJichev: I don't think this would work, since `Enum.drop([1, 2, 3], -1) # => [1, 2]` and you would lose the body data. https://hexdocs.pm/elixir/Enum.html#drop/2 I was thinking of doing it the same way, by using drop, but saw that it didn't return the dropped element.

whatyouhide: > (although you meant split :P)

:P

test/ex_unit_properties_test.exs
whatyouhide: @josevalim Elixir seems a bit too broad, but I am okay with adding it there. Do you have a preference?

josevalim: If the plan is inclusion in Elixir, then yes. Otherwise, no.

PR #85, opened 2018-02-09
map_of/3 accepts options instead of max_tries

davydog187:


lib/stream_data.ex
```diff
@@ -1164,12 +1164,14 @@ defmodule StreamData do
...
-  @spec map_of(t(key), t(value), non_neg_integer()) :: t(%{optional(key) => value})
+  @spec map_of(t(key), t(value), keyword()) :: t(%{optional(key) => value})
         when key: term(),
              value: term()
-  def map_of(key_data, value_data, max_tries \\ 10) do
+  def map_of(key_data, value_data, options \\ []) do
+    options = Keyword.put(options, :uniq_fun, fn {key, _value} -> key end)
+
     {key_data, value_data}
-    |> uniq_list_of(uniq_fun: fn {key, _value} -> key end, max_tries: max_tries)
```
davydog187: `uniq_list_of` already defaults `max_tries` to 10

lib/stream_data.ex
```diff
@@ -1164,12 +1164,14 @@ defmodule StreamData do
   Shrinks towards smallest maps and towards shrinking keys and values according
```
davydog187: I need to update the docs here to reflect the api change

lib/stream_data.ex
```diff
@@ -1156,20 +1156,22 @@ defmodule StreamData do
 
   ## Examples
 
-      Enum.take(StreamData.map_of(StreamData.integer(), StreamData.boolean()), 3)
+      Enum.take(StreamData.map_of(StreamData.integer(), StreamData.boolean()), max_tries: 3)
```
michalmuskala: The `3` argument was to `Enum.take/2`, so it was correct

davydog187: Good catch.

lib/stream_data.ex
whatyouhide: @davydog187 no worries, I quickly pushed an updated to the docs (and fixed the other comment from Michal). :)

whatyouhide:
Thanks a lot @davydog187, great job! 💟

PR #92, opened 2018-03-25
Emit tuples, maps and lists at equal frequency in term

josevalim:


lib/stream_data.ex
```diff
@@ -1868,7 +1868,18 @@ defmodule StreamData do
...
+    end)
+  end
+
+  defp one_to_four_element_tuple(leaf) do
+    bind(integer(), fn integer ->
+      case rem(integer, 10) do
+        rem when rem >= 6 -> {leaf, leaf, leaf}
+        rem when rem >= 3 -> {leaf, leaf}
+        rem when rem >= 1 -> {leaf}
+        _ -> {}
```
josevalim: We also emit empty tuples less frequently. (1/10)

lib/stream_data.ex
```diff
@@ -1868,7 +1868,18 @@ defmodule StreamData do
...
     tree(simple_term, fn leaf ->
-      one_of([list_of(leaf), map_of(leaf, leaf), {}, {leaf}, {leaf, leaf}, {leaf, leaf, leaf}])
+      one_of([list_of(leaf), map_of(leaf, leaf), one_to_four_element_tuple(leaf)])
+    end)
+  end
+
+  defp one_to_four_element_tuple(leaf) do
+    bind(integer(), fn integer ->
+      case rem(integer, 10) do
+        rem when rem >= 6 -> {leaf, leaf, leaf}
```
josevalim: Emit 3 element tuples more frequently. (4/10)

whatyouhide: Since you're binding on `integer()`, it means that the more the generation size grows, the more likely it will become to generate 3-element tuples. If you want to keep the percentages of `{}`, 1-element, 2-element, and 3-element tuples roughly constant, use `integer(0..10)` instead.

josevalim:
@whatyouhide updated. please check.

lib/stream_data.ex
michalmuskala: Why `rem` if the range of the integer is `0..9`?

PR #96, opened 2018-04-11
Fix positive_integer crashing for size = 0

obrok:
My main motivation for this is that I use something like the following to generate a typed expression tree:

```elixir
sized(fn size ->
  frequency([
    {1, leaf(type)},
    {size,  branch(type) |> resize(div(size, 2))}
  ])
end)
```

the size often becomes 0 in the leaves. The other generators seem to handle it well, but `positive_integer` crashes.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -1441,6 +1441,7 @@ defmodule StreamData do
   @spec positive_integer() :: t(pos_integer())
   def positive_integer() do
     new(fn seed, size ->
+      size = max(size, 1)
       {init, _next_seed} = uniform_in_range(1, size, seed)
       integer_lazy_tree(init, 1, size)
     end)
--- a/test/stream_data_test.exs
+++ b/test/stream_data_test.exs
@@ -230,9 +230,9 @@ defmodule StreamDataTest do
       end
     end
 
-    property "resized to 0" do
-      check all int <- resize(positive_integer(), 0) do
-        assert 1 = int
+    property "works when resized to 0" do
+      check all int <- resize(positive_integer(), 0), max_runs: 3 do
+        assert int == 1
       end
     end
   end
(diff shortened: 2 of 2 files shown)


obrok:
The travis failure seems to be unrelated?

josevalim:
@obrok yes, it is unrelated. :)

lib/stream_data.ex
```diff
@@ -1441,6 +1441,7 @@ defmodule StreamData do
   @spec positive_integer() :: t(pos_integer())
   def positive_integer() do
     new(fn seed, size ->
+      size = max(size, 1)
```
josevalim: Maybe a better fix would be to generate from 0 until size and then increment it by 1?

obrok: The problem is that `uniform_in_range(1, size, seed)` crashes when `size = 0`. I guess an alternative implementation would be `integer() |> map(&abs(&1) + 1)` which I'm using in my project for the time being. WDYT?

whatyouhide: @obrok that's slower and not much shorter, so I wouldn't go with it myself.

obrok: @whatyouhide so you're in favor of this version or do you think I should do something else?

whatyouhide:
I am not sure this is a good solution in general. I am not sure that we should ever let the size be 0, as it semantically doesn't really make sense. @josevalim thoughts?

josevalim:
@whatyouhide doesn't size 0 make sense for a list for example? if not, then we should enforce it to be > 0 in resize.

obrok:
The typespec currently says `@typep size() :: non_neg_integer()`. I was also wondering if maybe doing something like `size = size |> round() |> max(1)` somewhere like `resize` or `all` would be a valid alternative.

whatyouhide:
Thanks @obrok! 💟

PR #102, opened 2018-05-06
Pass in options correctly in do keyword syntax.

NJichev:
Doing a:
  `check all i <- integer(), max_runs: 25, do: assert ...`

failed with a match error because the options were packed with the body.

Fixed by extracting the options for a second time with `&Enum.split/2`.
Match on the body again to make sure that people don't pass in options
after the body.
Pack the options in another list since compile_check_all expects them
that way.

Added a test with a counter to verify this behaviour.

### DIFF
--- a/test/ex_unit_properties_test.exs
+++ b/test/ex_unit_properties_test.exs
@@ -188,7 +188,18 @@ defmodule ExUnitPropertiesTest do
       check all int1 <- integer(),
                 int2 <- integer(),
                 sum = abs(int1) + abs(int2),
+                max_runs: 25,
                 do: assert(sum >= int1)
+    end
+
+    test "do keyword syntax passes in options" do
+      {:ok, counter} = Agent.start_link(fn -> 0 end)
+
+      check all int <- integer(),
+                max_runs: 25,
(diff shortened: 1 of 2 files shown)


test/ex_unit_properties_test.exs
```diff
@@ -197,7 +197,7 @@ defmodule ExUnitPropertiesTest do
 
       check all int <- integer(),
                 max_runs: 25,
-                do: Agent.update(counter, &(&1 + 1)) && assert is_integer(int)
```
NJichev: Ehhh, wasn't assert in the default .formatter.exs.

whatyouhide: It is but we always parenthesize in some cases (I think this case is because the `assert` call is part of the `&&` call.

harlantwood:
Just mentioning in case it went unnoticed: looks like this PR broke the travis build: https://travis-ci.org/whatyouhide/stream_data/builds/381172960

@NJichev @whatyouhide

NJichev:
That's not part of the commit, the test that failed filters 5's, sometimes the generator will just generate too many of those values that leads to this error :)

PR #104, opened 2018-06-04
Add StreamData.repeatedly to pass in a function as your generator

aforward:
This PR is based on the conversation at:

https://elixirforum.com/t/how-to-create-a-custom-streamdata-generator/11935/9

And a similar need where I wanted to test the properties of the results of a function call.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -251,6 +251,23 @@ defmodule StreamData do
     new(fn _seed, _size -> lazy_tree_constant(term) end)
   end
 
+  @doc """
+  A generator that always generates a value by calling `fun`.
+
+  ## Examples
+
+      iex> Enum.take(StreamData.repeatedly(fn -> :rand.uniform() end), 2)
+      [0.9933462977419845, 0.0951497908766279]
+
+  ## Shrinking
+
+  This generator doesn't shrink.
+  """
+  @spec repeatedly((() -> a)) :: t(a) when a: var
+  def repeatedly(fun) when is_function(fun, 0) do
+    new(fn _seed, _size -> lazy_tree_constant(fun.()) end)
+  end
+
   ## Combinators
 
   @compile {:inline, map: 2}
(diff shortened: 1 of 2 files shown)


whatyouhide:
The problem with this is that is defeats the purpose of StreamData. StreamData is there so you can generate random stuff that has nice properties (shrinks, varies size). This generator would defeat the purpose.

The use case mentioned in ElixirForum doesn't seem a good use case because if you want to test the results of a function, you don't need StreamData:

```elixir
for _ <- 1..100 do
  {:ok, token} = AuthToken.generate_key()
  # test properties here
end
```

StreamData will give you nothing over this if you use `repeatedly`.

Can you give me a concrete use case of `repeatedly`? So I can see if this is actually needed or if what I said above makes sense 😄

aforward:
The use case is to leverage the other macros with the property testing so that it reads nicely

```elixir
  property "nonce is always unique" do
    check all nonces <- nonempty(list_of(nonce())) do
      assert Enum.uniq(nonces) == nonces
    end
  end

  property "nonce is always verifiable" do
    check all nonce <- nonce() do
      assert :ok == AcmeServer.Nonce.verify!(nonce)
    end
  end

  property "nonce can only be verified once" do
    check all nonce <- nonce() do
      :ok = AcmeServer.Nonce.verify!(nonce)
      assert_raise MatchError, fn -> AcmeServer.Nonce.verify!(nonce) end
    end
  end

  defp nonce() do
    # we need to map a constant to ensure we get a different nonce every time
    unshrinkable(map(constant(:ignore), fn _ -> AcmeServer.Nonce.new() end))
  end
```

Here `AcmeServer.Nonce.new()` already provides random data that we want to assert certain propeties about it.  

The implementation of `unshrinkable(map(constant(:ignore), fn _ -> AcmeServer.Nonce.new() end))` to achieve this (or the one from the elixir forum) seem a little obtuse to leverage the other great capabilities of this library, namely `property` and `check all`.

whatyouhide:
What you're doing is 99% doable with just a `for` comprehension, really no need for property-based testing. Shrinking wouldn't work so I think this promotes something we'd rather not promote. Thanks a lot for the contribution though! 💟

PR #108, opened 2018-07-23
Fixed example in docs for fixed_map/1

lasseebert:
In the docs for `fixed_map`, the `:int` key in the example result should be `:integer`.


whatyouhide:
Thanks @lasseebert! 💟

Issue #106, opened 2018-07-17
Parallel properties


jjl:
It would be nice to be able to run a single property in parallel in order that the full run might finish quicker.

I have a complex property that starts to take a very long time towards the end of the run (diagnosed by adding `IO.inspect(:ok)` to the end of each run). I'm able to throw cores at the problem, but currently, StreamData isn't capable of using them.

Is there interest in this? Would a contribution to add this be welcome?

luc-tielen:
This would be a cool feature imo.

whatyouhide:
@jjl yes, this is a good way of improving StreamData. It was proposed before to do this by parallelizing properties (instead of inside a single property), but that would require major changes to ExUnit, so I would consider this solution better. @jjl would you have time to experiment with this and send out a proof of concept PR? We should likely use `Task.async_stream` for this, and start shrinking on the first error that we get (basically pipe into `Enum.find` or something like that to stop the stream). Let me know if you would like to work on this.

jjl:
I'll see what I can do about it, but time is a little constrained at the minute, so more likely to be later than now.

jjl:
I think the easiest way would be to make `check all` wrap the body in an extra function before passing to `gen all` and then do a `Task.async_stream(fns, fn x -> x.())`.

Does this sound like a reasonable approach?

whatyouhide:
@jjl yes, that sounds about right. :)

whatyouhide:
Closing in favour of the PR (#112).

PR #112, opened 2018-10-07
Parallel properties

whatyouhide:
Related to #106.

@josevalim opening this mostly to get your feedback. To make parallel properties work, I had to change `check_all` to create an enumerable that returns lazy trees so that we can do map or async_stream on it, depending on `:parallel`. The main problem is that while this works, I think in many many cases it will make things much slower because of all the processes that are created. This is really only useful if the heavy work you do is in the `do`/`end` block you pass to `check all` as it won't make generating happen in parallel. THoughts? Does it make the code too complex and it's not worth it?

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -1991,6 +1991,7 @@ defmodule StreamData do
     seed = new_seed(Keyword.fetch!(options, :initial_seed))
     size = Keyword.get(options, :initial_size, 1)
     max_shrinking_steps = Keyword.get(options, :max_shrinking_steps, 100)
+    parallel? = Keyword.get(options, :parallel, false)
     start_time = System.system_time(:millisecond)
     config = %{max_shrinking_steps: max_shrinking_steps}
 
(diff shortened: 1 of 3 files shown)


whatyouhide:
Now that I think about it, we *might* be able to generate the next term in parallel if we basically generate a stream of random seeds and get the next term from the generator inside the async_stream function... 🤔

lib/stream_data.ex
```diff
@@ -2014,34 +2015,77 @@ defmodule StreamData do
           Map.merge(config, %{max_end_time: start_time + max_run_time, max_runs: max_runs})
       end
 
-    check_all(data, seed, size, fun, _runs = 0, start_time, config)
+    data
+    |> generator_to_enum(seed, size, start_time, config)
```
josevalim: I think you can use `Stream.transform` instead of this function, no?

whatyouhide:
@josevalim I reworked this to use just a bunch of streams functions, and it looks nice, but this benchmark still shows that the parallel one suffers *a lot* when the generators generate complex trees (which are basically just a ton of nested anon functions):

```elixir
Benchee.run(%{
  "check_all par" => fn ->
    check all strings <- list_of(string(:printable)), parallel: true, max_runs: 100 do
      assert is_list(strings)
    end
  end,
  "check_all seq" => fn ->
    check all strings <- list_of(string(:printable)), parallel: false, max_runs: 100 do
      assert is_list(strings)
    end
  end
})
```

jjl:
Random number generation in parallel could be done with some sort of splittable randomness source (see e.g. https://stackoverflow.com/a/47523176/828125 for a simple explanation of splittable randomness if you haven't come across it before). 

*However*, this means that the repeatability becomes tied to the number of schedulers the suite runs with erlang configured to use. There's a hack whereby you impose a limit of say 128 and reuse them, but it's a bit fugly.

whatyouhide:
@jjl Number generation is fast so we don't need to do it in parallel. The current solution generates a stream of lazy trees and then runs the function passed to `check_all` in parallel. A previous solution used a stream of random seeds and generated the lazy tree + called the function passed to `check_all` in parallel, but they both show performance problems

jjl:
Ah, right, I misunderstood.

The approach I've taken with a library I've been working on to get around the performance problems of all the nested functions is to build a vector of structs representing instructions and then have an evil macro that evals the code that builds them and generates a module containing optimised code.

It's a bit mad, but the benchmarks are pretty good compared to the interpreted version.

whatyouhide:
@jjl yep, that's a fairly common-ish solution to this kind of problem. [nimble_parsec](https://github.com/plataformatec/nimble_parsec) does this. However, it would mean a pretty big restructuring of all of stream_data and that's not something I plan on doing anytime soon. If you want to experiment with it, I'll be happy to take a look though :)

whatyouhide:
Oh well, closing this for now. :)

Issue #117, opened 2018-12-29
make initial_seed overridable


isaac-rstor:
currently, you can't override initial_seed during property checking

I've got a scenario where I'm generating random modules in two different test sets, and the module names are colliding, which is an annoyance (tons of "modules redefined" warnings) when in synchronous testing and a serious problem when in sync mode. 

on line 440:

https://github.com/whatyouhide/stream_data/blob/master/lib/ex_unit_properties.ex#L440

we see the any attempt to set a seed get thrown away.

I'll PR a change to this, including documentation.

PR #118, opened 2018-12-30
Add initial seed option

isaac-rstor:
addresses 

https://github.com/whatyouhide/stream_data/issues/117

adds documentation, and also some quick unit tests.  There appears to be a TODO in the block, indicating that more work is projected for this section, the TODO has been preserved and placed ahead of the new code.

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -432,12 +432,17 @@ defmodule ExUnitProperties do
     quote do
       options = unquote(options)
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
Thanks for the great work @isaac-rstor! I slightly simplified the code and pushed a commit that implements this (dd95ec2). I also removed support for three-element tuples as it's something we can always add in the future. I think this option is good to have but I'd prefer that users go with ExUnit's seed since in most cases that's the best choice. Anyways, thanks a lot for your contribution!

PR #125, opened 2019-03-29
Add documentation for changing `max_runs` via config

TylerPachal:
It was non-obvious how I could set the `max_runs` via my config files so that I could change the value globally depending on where my tests were running (I wanted a lot of runs on CircleCI, and a few runs on my local machine).  I ended up figuring how, don't think it was documented anywhere.

Here is what it looks like (new part at the bottom):

Screen Shot 2019-04-01 at 11 47 33 AM


whatyouhide:
The change is great, but I think we should move this to the documentation for `ExUnitProperties` since that's where users of property-based testing and `check all` should look.

TylerPachal:
@whatyouhide How about that? (Updated the screenshot at the top of the thread)

whatyouhide:
Thanks @TylerPachal! 💟

PR #132, opened 2019-06-27
Add "see also" for map functions and bump ex_doc version

axelson:
I found `optional_map` in the docs, but I nearly missed `fixed_map`. It's easier in the source since they're next to each other, but in the docs they're sorted alphabetically so it's a little bit harder to find. I also bumped the ex_doc version.


whatyouhide:
Thanks @axelson! 💟

axelson:
Thanks for merging!

PR #133, opened 2019-06-29
Typespec fixes, fix #130

### LINKED ISSUE #130: Generators make Dialyzer complain that a function has no local return
Given this code:

```elixir
  @spec optional_alphanum(pos_integer) :: %StreamData{}
  def optional_alphanum(size) when size > 0 do
    gen all str <- string([?A..?Z, ?0..?9], min_length: 0, max_length: size) do
      String.pad_trailing(str, size, " ")
    end
  end
```

Dialyzer says: `Function optional_alphanum/1 has no local return`. I also tried appending `| no_return()` to the return type, and it did not help.

Can we make Dialyzer happy with `ExUnitProperties` / `StreamData` -- and how?

(Copied from https://elixirforum.com/t/property-generators-make-dialyzer-complain-ideas-how-to-make-it-happy/23049)

rodrigues:
About changing `@opaque` for `@type`, not really sure if this should change, but it fixes the following dialyzer error:
`{:warn_contract_types, {'lib/ex_unit_properties.ex', 605}, {:invalid_contract, [ExUnitProperties, :pick, 1, '(atom() | tuple() | \#{\'__struct__\':=\'Elixir.StreamData\', \'generator\':=fun((_,_) -> any()), _=>_}) -> any()']}}`

This still doesn't fix https://github.com/whatyouhide/stream_data/issues/130 's `no return`, but I have one less error related to stream_data, the following one is gone from [my project](https://github.com/rodrigues/deli/blob/master/test/support/stream_generators.ex):

```
test/support/stream_generators.ex:38:call_without_opaque
Function call without opaqueness type mismatch.

Call does not have expected opaque term of type StreamData.t(_) in the 1st position.

StreamData.nonempty(%StreamData{:generator => (_, _ -> map())})
```

EDIT: other commits were added to this PR, and it now fixes #130. More about it in the following comment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ elixir: '1.9'
 jobs:
   - check formatted
   - test
+  - check typespecs
 
 script:
   - mix test --include stdlib
@@ -28,3 +29,10 @@ jobs:
 
     - stage: check formatted
       script: mix format --check-formatted
+
+    - stage: check typespecs
+      script: MIX_ENV=dev mix do clean, compile, dialyzer --halt-exit-status
+
+cache:
+  directories:
+  - _build/dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,6 @@ jobs:
   include:
     - stage: test
 
-    - otp_release: '18.3'
-      elixir: '1.5'
-
     - otp_release: '19.3'
       elixir: '1.6'
 
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -107,10 +107,15 @@ defmodule StreamData do
   @typep generator_fun(a) :: (seed(), size() -> LazyTree.t(a))
 
   @typedoc """
+  Valid input to create a `StreamData.t()` type
+  """
+  @type input(a) :: t(a) | atom() | tuple()
+
+  @typedoc """
   A type that represents a `StreamData` generator that generates values
   of type `a`.
   """
-  @type t(a) :: %__MODULE__{generator: generator_fun(a)} | atom() | tuple()
+  @type t(a) :: %__MODULE__{generator: generator_fun(a)}
 
   # TODO: remove once we depend on OTP 20+ since :exs64 is deprecated.
   if String.to_integer(System.otp_release()) >= 20 do
@@ -275,7 +280,7 @@ defmodule StreamData do
   This generator shrinks exactly like `data`, but with `fun` mapped over the
   shrunk data.
   """
-  @spec map(t(a), (a -> b)) :: t(b) when a: term(), b: term()
+  @spec map(input(a), (a -> b)) :: t(b) when a: term(), b: term()
   def map(data, fun) when is_function(fun, 1) do
     new(fn seed, size ->
       data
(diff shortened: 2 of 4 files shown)


rodrigues:
This still doesn't fix https://github.com/whatyouhide/stream_data/issues/130 's `no return`, but I have one less error related to stream_data, the following one is gone from my project:

```
test/support/stream_generators.ex:38:call_without_opaque
Function call without opaqueness type mismatch.

Call does not have expected opaque term of type StreamData.t(_) in the 1st position.

StreamData.nonempty(%StreamData{:generator => (_, _ -> map())})
```

rodrigues:
This commit above fixes the dialyzer error that happens a lot [(full log)](https://gist.github.com/rodrigues/a7f79f8f03e42e3ed0a7b201be7924e9):

Here compacted:

```
$ mix dialyzer
…
lib/stream_data.ex:252:extra_range
Type specification has too many types.

Function:
StreamData.constant/1

Extra type:
atom() | tuple()

Success typing:
%StreamData{:generator => (_, _ -> map())}
________________________________________________________________________________
(…same error quite a few times…)
```

Now the only dialyzer missing after this PR fixes is also the worst one:

```
lib/stream_data.ex:344:call
The function call will not succeed.

StreamData.bind_filter(_data :: any(), (_, _ -> any()), _max_consecutive_failures :: any())

breaks the contract
(t(97), (97 -> {:cont, t(98)} | :skip), non_neg_integer()) :: t(98)
when a: term(), b: term()
```

whatyouhide:
Hey @rodrigues, thanks so much for the contribution. I am on the fence on this one. I mostly use `@spec`s for documentation, and I think this change makes the documentation more obscure. I'm gonna have to think about this.

rodrigues:
@whatyouhide Sure man, take your time. Let me know if you want me to change something.

whatyouhide:
Hey @rodrigues, after some thought, I’m not comfortable with reducing the clarity of the typespecs (which I want to mostly use for documentation) in order to fix something that I *think* should work. Also, I think it's pretty rare that you would either run Dialyzer on your test suite **or** use stream_data in production code, so I'll close this. Thanks so much for the contribution though, very appreciated! 💟

rodrigues:
Hi @whatyouhide, I understand your perspective, but as you don’t seem to use dialyzer very much, I’m bringing also a bit of my thoughts on it, if it helps.

For codebases that use dialyzer, it makes a lot of sense to run it with the tests included too, after all it’s code that actually calls the functions with “real” usage examples, specially when using generators like stream_data, it’s good to make sure you’re creating valid test cases. Making stream_data not dialyzer-friendly might make dialyzer users to need to use something else.

I’d prefer to make it dialyzer-friendly, and address the documentation issues, but of course, I understand where you’re coming from, specially given current dialyzer adoption among Elixir devs.

Cheers!

whatyouhide:
> For codebases that use dialyzer, it makes a lot of sense to run it with the tests included too, after all it’s code that actually calls the functions with “real” usage examples, specially when using generators like stream_data, it’s good to make sure you’re creating valid test cases.

Fair enough.

> I’d prefer to make it dialyzer-friendly, and address the documentation issues

The problem here is that this fix makes the documentation noticeably more difficult to understand in order to fix a tool. I believe that the original implementation of the specs should have worked and in fact I don't really get why this patch fixes things. I'm definitely open to fixing the Dialyzer warnings, but I wanna do it not at the cost of documentation :)

PR #135, opened 2019-10-28
Fix typos in docs

preciz:
Thanks for reviewing :wave: .


lib/stream_data.ex
```diff
@@ -698,7 +698,7 @@ defmodule StreamData do
 
   ## Examples
 
-  Let's build a generator that returns a binary around 25% of times and a
+  Let's build a generator that returns a binary around 25% of times and an
```
bryanhuntesl: *of the time and an integer around 75% of the time.*

bryanhuntesl: everything else is good

preciz: Thanks, I have updated it. :+1:

whatyouhide:
Thank you @preciz 💟

PR #140, opened 2020-04-19
Fixed typo in docs of list_of

elbow-jason:
Added a closing square bracket to an unclosed list in one of the examples in the docs of `list_of`


whatyouhide:
Great catch, thanks! 💟

PR #142, opened 2020-06-23
Add recent Erlang/OTP and Elixir versions to CI

adrianomitre:
Main contribution:
- Add recent Erlang/OTP and Elixir versions to CI. Please refer to https://github.com/dashbitco/broadway/pull/188 for justification for minimum and maximum Erlang/OTP for each minor Elixir (last patch) version vs. only minimum of both and maximum of both.

Additional ones:
- Upgraded GitHub actions to latest versions
- Only get test dependencies on test job
- Compile with MIX_ENV=test
- Treat warnings as errors when compiling on last Elixir version


adrianomitre:
The difference in results in the following two test runs on the exact same code indicates that [`test/ex_unit_properties_test.exs:6`](https://github.com/whatyouhide/stream_data/blob/master/test/ex_unit_properties_test.exs#L6) is non deterministic (at least on Elixir 1.6.6):
- https://github.com/whatyouhide/stream_data/pull/142/checks?check_run_id=799325492
- https://github.com/whatyouhide/stream_data/pull/142/checks?check_run_id=799338011

.github/workflows/main.yml
```diff
@@ -27,30 +27,57 @@ jobs:
...
+        if: matrix.elixir != '1.10.3'
+        run: mix compile
+        env:
+          MIX_ENV: test
+
+      - name: Compile application treating warnings as errors (only for last Elixir version)
+        if: matrix.elixir == '1.10.3'
+        run: mix compile --warnings-as-errors
+        env:
+          MIX_ENV: test
```
whatyouhide, line 80: Can't we make this as another task, like we do formatting? That is, explicitly add a task for compiling with warning on latest Erlang/Elixir.

adrianomitre: Sure can do, but why do you believe that would be better?

adrianomitre:
[Line 11 of `mix.exs` reads ](https://github.com/whatyouhide/stream_data/blob/master/mix.exs#L11) `elixir: "~> 1.5",` but the tests [`test/stdlib/string_test.exs:8`](https://github.com/whatyouhide/stream_data/blob/master/test/stdlib/string_test.exs#L8) are failing on Elixir 1.5.3.

.github/workflows/main.yml
whatyouhide: Because I'm not a fan of the `if`ing around. I'd prefer "Compile with warnings" to be a different task. Thanks!

adrianomitre: A little less DRY, but the `if`s are gone, as requested: https://github.com/whatyouhide/stream_data/pull/142/commits/a5682fd1770a367bee96d4252f618bee5316279d.

whatyouhide:
Removed a few versions that we don't need to check, but looks great like this, thanks!

whatyouhide:
Removed a few versions that we don't need to check, but looks great like this, thanks!

PR #146, opened 2020-12-05
Make pick/1 generation size configurable

garthk:
A solution to a self-inflicted problem discovered while using StreamData to generate replies for `Mox` pretending to be `Elasticsearch.API`.

Let's say you got carried away with the realism of your data generation:

```elixir
import StreamData

word = string(?a..?z, min_length: 1, max_length: 12)
words = list_of(word)
sentence = map(words, fn a -> [Enum.intersperse(a, " "), "."] end)
paragraph = map(list_of(sentence), fn a -> [Enum.intersperse(a, " ")] end)
document = map(list_of(paragraph) , fn a -> [Enum.intersperse(a, "\n\n")] end)

test_doc = ExUnitProperties.pick(document)
```

Given a particularly unlucky seed, generating such a test document can take five seconds on a 3.2 GHz Xeon W, causing your tests to fail:

```elixir
:rand.seed({:exs64, 10998329948955527640})
{us, iodata} = :timer.tc(&ExUnitProperties.pick/1, [document])
{us / 1_000_000, IO.iodata_length(iodata) / 1_048_576}
```

To save wading through the code sprinkling `resize/1` everywhere, this change gives you a quick way to limit the scale of your problem:

```elixir
config :stream_data, pick_generation_size: 1..5
```


garthk:
D'oh! Forgot to commit the `application.env` changes in `mix.exs`; I'd added `dialyxir` for some extra safety and over-did avoiding adding it to this PR :)

whatyouhide:
A mere 2.5y later I’m on this! You should be able to resize the top-level generator, and it will propagate to all the "inner" generators 🙃

PR #148, opened 2021-01-25
Feature: Not implemented property

endersstocker:
Why
---

When I test a given module or function, I want to write a not
implemented property, so that I can remember to implement it later as I
do with **ExUnit.Case.test/1**

What
----

Add **property/1** to **ExUnitProperties**

Screen Shot 2021-01-24 at 8 01 26 PM

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -114,6 +114,27 @@ defmodule ExUnitProperties do
   end
 
   @doc """
+  Defines a not implemented property test with a string.
+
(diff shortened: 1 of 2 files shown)


endersstocker:
The new test is the only one in **ExUnitPropertiesTest** that requires synchronous execution. Performance of the test suite could be optimized by moving this test into its own case, at the expense of cohesion.

whatyouhide:
Nice addition, thanks @endersstocker!

PR #151, opened 2021-02-11
Fix typo in the docs for optional_map

holsee:
Fixed the key in the example result to match the parameters within the example.


whatyouhide:
Good catch, thanks! 💟

PR #153, opened 2021-06-24
add-collect-option

TylerPachal:
Experiment to add new `:collect` option which will print the distribution of values produced by a generator.  This is inspired by the [`collect/2` and `aggregate/2` functions in PropEr](https://proper-testing.github.io/apidocs/proper.html#collect-2).

This is just a first attempt as an experiment to see if it was possible.  Do you folks think something like this would be useful?  It is probably the feature I miss the most from PropEr.  I am completely open to changing the interface, the output, etc.  This is just meant to get a discussion going.

---

Example test 1:

```
property "collect test 1" do
  check all b <- boolean(), [collect: :b] do
    assert is_boolean(b)
  end
end
```

Output
```
$ stream_data git:(add-collect-option) mix test test/ex_unit_properties_test.exs:285
Excluding tags: [:test]
Including tags: [line: "285"]

Collect:
    45.0% true
    55.0% false

.

Finished in 1.0 seconds
11 properties, 12 tests, 0 failures, 22 excluded

Randomized with seed 577891
```

---

Example test 2

```
property "collect test 2" do
  to_range = fn (i) ->
    base = div(i, 10)
    {base * 10, (base + 1) * 10}
  end

  check all i <- integer(0..100), [collect: {:i, to_range}] do
    assert is_integer(i)
  end
end
```

Output
```
$ stream_data git:(add-collect-option) ✗ mix test test/ex_unit_properties_test.exs:285
Excluding tags: [:test]
Including tags: [line: "285"]

Collect:
    15.0% {40, 50}
    15.0% {60, 70}
    15.0% {70, 80}
    10.0% {0, 10}
    10.0% {80, 90}
    9.0% {20, 30}
    8.0% {50, 60}
    7.0% {10, 20}
    6.0% {90, 100}
    5.0% {30, 40}

.

Finished in 1.0 seconds
11 properties, 12 tests, 0 failures, 22 excluded

Randomized with seed 827184
```

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -522,9 +522,16 @@ defmodule ExUnitProperties do
                 {:error, result}
             else
               _result ->
-                with {:ok, collectable_name} <- Keyword.fetch(unquote(options), :collect),
+                collect_name =
+                  case Keyword.fetch(unquote(options), :collect) do
+                    {:ok, {collect_name, _func}} -> {:ok, collect_name}
+                    {:ok, collect_name} -> {:ok, collect_name}
+                    _ -> :error
+                  end
+
+                with {:ok, collect_name} <- collect_name,
                      generated_values = var!(generated_values, unquote(__MODULE__)),
-                     {_clause, value} <- Enum.find(generated_values, fn {clause, _value} -> String.starts_with?(clause, "#{collectable_name} <-") end)
+                     {_clause, value} <- Enum.find(generated_values, fn {clause, _value} -> String.starts_with?(clause, "#{collect_name} <-") end)
                 do
                   {:ok, %{collect: value}}
                 else
@@ -618,11 +625,11 @@ defmodule ExUnitProperties do
 
     distribution =
       distribution
-      |> Enum.sort_by(fn {_val, count} -> count end)
+      |> Enum.sort_by(fn {_val, count} -> count end, :desc)
       |> Enum.map_join("\n", fn {val, count} ->
         percentage = :erlang.float_to_binary(count / total * 100, decimals: 1)
         indent(String.trim_trailing("""
-        #{percentage}% #{val}
+        #{percentage}% #{inspect(val)}
         """), "    ")
       end)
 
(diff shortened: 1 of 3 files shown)


whatyouhide:
Hey @TylerPachal, hi! 👋 I’m finally having a moment to look into this. Thanks in advance for taking the time and tackling this.  Leaving some thoughts below. In general, I like this, always wanted this, and maybe now's the time 🙃 

## API

The most important thing to figure out for me is the API. I have the following concerns:

1. Can we find a way to not _print_ this automatically for every test, but rather make that plug-in? That is, I want to be able to collect and then I want users to be able to do whatever they want with the collected data. Essentially, split the "collection" part into just collecting into a data structure and have a separate "printing" phase where that data structure is given to something else (like user-provided printers). I don't want to come up with a complex and pluggable API right away but I want to get the design of this right early on.
2. Should we collect *per generator*, like you did, or potentially have a "global" collection for all the generated data?

```elixir
check all ...,
          collect: fn -> data_structure_here end do
end
```

## Other frameworks

Other than PropEr, can we research other property-based testing frameworks and see how they implement collection? I’m particularly interested in [Clojure's test.check](https://github.com/clojure/test.check), which has been by far the strongest inspiration for stream_data.

TylerPachal:
Thanks for your thoughts.

> Can we find a way to not print this automatically for every test.

I believe PropEr does this, where you need to hand in the `:verbose` flag combined with the actual `collect` call, though it is still pretty restrictive in its actual printing:

```
# propcheck example
#
# example output
#
# 15.00% 9
# 14.00% 2
# 13.00% 7
# 12.00% 4
# 10.00% 8
# 9.00% 3
# 8.00% 5
# 7.00% 0
# 7.00% 1
# 5.00% 6
#
property "collect/2 example", [:verbose] do
  forall i <- integer(0, 1000) do
    assertion = i >= 0
    collect_value = div(i, 100)
    collect(assertion, collect_value)
  end
end
```

---

I will check out the Clojure library and see what I can come up with.

PR #154, opened 2021-07-04
Replace setup-elixir with setup-beam

nallwhy:
[setup-elixir](https://github.com/actions/setup-elixir) is unmaintained and it's now maintained at [setup-beam](https://github.com/erlef/setup-beam).

This PR replaces `setup-elixir` with `setup-beam` and change version matrix for test(adds OPT 24 and elixir 1.12.2 & removes OTP 19 because `setup-beam` support OTP >= 20 on ubuntu-latest).


whatyouhide:
Ah, thank you for taking care of this! 💟

PR #156, opened 2021-11-02
Implementation of `StreamData.non_negative_integer/0` generator

Qqwy:
This PR adds `StreamData.non_negative_integer()` which is a rather common type to be e.g. used when generating 'sizes'.

Without it, we either would be limited to a particular maximum (using the `StreamData.integer(0..size)` syntax) or had to resort to hacks which are more verbose and slower `StreamData.integer() |> StreamData.map(&abs/1)`.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -1514,6 +1514,27 @@ defmodule StreamData do
   end
 
   @doc """
+  Generates non-negative integers bound by the generation size.
+
+  ## Examples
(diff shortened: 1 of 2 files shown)


Qqwy:
I think this is the most common generator which is still missing.

If desired, two other generators which are slightly less prevalent but still common enough to warrant a shorthand status might be `negative_integer()` and `number()`.

PR #159, opened 2021-12-21
Implementation of StreamData.integer(range) for ranges with steps.

### LINKED ISSUE #158: StreamData.integer/1 does not handle ranges with steps correctly
Given

```elixir
generator = StreamData.integer(2..8//2)
```

When calling 

```elixir
generator |> Enum.take(1_000) |> MapSet.new
```

the expected output is

```elixir
#MapSet<[2, 4, 6, 8]>
```
(i.e. the same as calling `MapSet.new(2..8//2)`)

but instead we get

```elixir
#MapSet<[2, 3, 4, 5, 6, 7, 8]>
```

It seems StreamData is not taking the `step` field (introduced in Elixir 1.12) into account.

Qqwy:
Fixes #158

- [x] Implementation for ranges with steps
- [x] Ensure backwards-compatibility using conditional compilation
- [x] New property-tests to test the new functionality on ranges with both positive and negative steps.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -505,14 +505,43 @@ defmodule StreamData do
   Shrinks towards the smallest absolute value that still lie in `range`.
   """
   @spec integer(Range.t()) :: t(integer())
-  def integer(left..right = _range) do
-    {lower, upper} = order(left, right)
+  # Range step syntax was introduced in Elixir v1.12.0
+  if Version.compare(System.version(), "1.12.0") == :lt do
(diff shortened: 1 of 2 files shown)


Qqwy:
@whatyouhide I believe the PR is done. 
There are still some CI failures but they don't seem to be related to the code I have added.

whatyouhide (approved):

whatyouhide:
Thanks @Qqwy! 💟

Issue #129, opened 2019-05-24
Generators for Date, Time, and DateTime values


whatyouhide:

Qqwy:
It's a good call to split my earlier issue into two. :+1: 

Just to make sure it does not get lost, you can see my earlier thoughts as to what functionality Date/Time/DateTime generator(s) might require [here](https://github.com/whatyouhide/stream_data/issues/126#issuecomment-486051843).

whatyouhide:
After a couple of years, I still think that having a good date generator is worth it but I'm afraid of putting it into the library because I think the requirements for what dates should look like vary a lot between applications. For now, unless someone wants to give this a shot, I'm closing this!

Qqwy:
While I agree that requirements for dates vary between applications, I think that there is a large majority of libraries and applications for which it is good enough to be able to generate:

- Any (future and past) (date)(time);
- Any future (date)(time);
- Any past (date)(time).

In some cases, it might make sense to round them to the nearest second/minute/hour/day, or to rescale them to ensure a larger range. Both of these things can be done using pre-existing tooling outside of the (date)(time) generators themselves (e.g. the tools of the Elixir standard library and/or Timex, and the use of `StreamData.scale/2`).

---

I am willing to give this a shot. Expect a PR in the next couple of days. :-)

PR #161, opened 2021-12-30
WIP: Date time generators

Qqwy:
Implementation of #129 

- [x] Date generator
  - [x] Tests
- [ ] Time generator
  - [ ] Tests
- [ ] DateTime generator
  - [ ] Tests
- [ ] NaiveDateTime generator
  - [ ] Tests

---

So far a generator function has added to create `Date` structs, including the accompanying tests.

The implementation for `Time`, `DateTime` and `NaiveDateTime` will be similar, except that their unit of offset will be fractions of milliseconds (rather than days), and of course the `Date.Range`-related code does not apply.

Since their implementation will be very similar, this seemed like a good moment to give you the opportunity to provide initial feedback on the code.

Happy holidays! :fireworks:

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -1922,60 +1922,84 @@ defmodule StreamData do
   """
   @spec date(Date.Range.t() | keyword()) :: t(Date.t())
   def date(options_or_date_range \\ [])
+
   def date(date_range = %Date.Range{}) do
-    # TODO step
-    date_between_bounds(date_range.first, date_range.last, date_range.first.calendar)
+    member_of(date_range)
   end
+
   def date(options) when is_list(options) do
     min = Access.get(options, :min, nil)
     max = Access.get(options, :max, nil)
     origin = Access.get(options, :origin, Date.utc_today())
+
     case {min, max} do
       {nil, nil} ->
         # TODO any date
         any_date(origin, origin.calendar)
+
       {nil, max = %Date{}} ->
         past_date(max, max.calendar)
+
       {min = %Date{}, nil} ->
         future_date(min, min.calendar)
+
       {min = %Date{}, max = %Date{}} ->
         if min.calendar != max.calendar do
-          raise ArgumentError, "Two dates with incompatible calendars were passed to `StreamData.date/1`"
+          raise ArgumentError,
+                "Two dates with incompatible calendars were passed to `StreamData.date/1`"
         end
+
         date_between_bounds(min, max, min.calendar)
     end
   end
 
   defp any_date(origin \\ Date.utc_today(), calendar \\ Calendar.ISO) do
-    {iso_days, day_fraction} = calendar.naive_datetime_to_iso_days(origin.year, origin.month, origin.day, 0, 0, 0, {0, 0})
+    {iso_days, day_fraction} =
+      calendar.naive_datetime_to_iso_days(origin.year, origin.month, origin.day, 0, 0, 0, {0, 0})
+
     StreamData.map(integer(), fn offset ->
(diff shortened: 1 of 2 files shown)


Qqwy:
@whatyouhide Just a gentle reminder in case the notification of this PR slipped through the cracks.

Of course, take your time, and happy new-year's holidays! :blush:

lib/stream_data.ex
```diff
@@ -1915,6 +1916,122 @@ defmodule StreamData do
     end)
   end
 
+  @doc """
+  Generates arbitrary (past and future) dates.
+
+  Generated values shrink towards `Date.utc_today/0`.
+  """
+  def date do
```
whatyouhide, line 1924:
```suggestion
  @spec date() :: t(Date.t())
  def date do
```

lib/stream_data.ex
```diff
@@ -1915,6 +1916,122 @@ defmodule StreamData do
...
+  * `:origin` - (`Date`) if present, generated values will shrink towards this date. Cannot be combined with `:min` or `:max`.
+
+  * `:min` - (`Date`) if present, only dates _after_ this date will be generated. Values will shrink towards this date.
+
+  * `:max` - (`Date`) if present, only dates _before_ this date will be generated. Values will shrink towards this date.
+
+  If both `:min` and `:max` are provided, dates between the two mentioned dates will be generated.
+  Values will shrink towards `:min`.
+
+  If no options are provided, will work just like `StreamData.date/0`.
```
whatyouhide, line 1942:
```suggestion
  If no options are provided, will work just like `date/0`.
```

lib/stream_data.ex
```diff
@@ -1915,6 +1916,122 @@ defmodule StreamData do
...
+
+  ## Date.Range
```
whatyouhide, line 1944:
```suggestion
  ## `Date.Range`
```

lib/stream_data.ex
```diff
@@ -1915,6 +1916,122 @@ defmodule StreamData do
...
+  Alternatively a `Date.Range` can be given. This will generate dates in the given range,
+  and with the supplied `date_range.step`.
+
+  Values will shrink towards `date_range.first`.
+
+  ## Calendar support
+
+  This generator works with `Calendar.ISO` and any other calendar
+  which implements the callbacks
+  `c:naive_datetime_to_iso_days/7` and `c:naive_datetime_from_iso_days/2`.
```
whatyouhide, line 1955:
```suggestion
  `c:Calendar.naive_datetime_to_iso_days/7` and `c:Calendar.naive_datetime_from_iso_days/2`.
```

lib/stream_data.ex
```diff
@@ -1915,6 +1916,122 @@ defmodule StreamData do
...
+      {nil, max = %Date{}} ->
+        past_date(max, max.calendar)
+
+      {min = %Date{}, nil} ->
+        future_date(min, min.calendar)
+
+      {min = %Date{}, max = %Date{}} ->
+        if min.calendar != max.calendar do
+          raise ArgumentError,
+                "Two dates with different calendars were passed to `StreamData.date/1`"
```
whatyouhide, line 1982:
```suggestion
                "dates with different calendars were passed to StreamData.date/1"
```

whatyouhide:
I like where this is going. Is there any way we can use `Date.add/2` (and eventually `DateTime.add/3` and friends) here instead of converting between days and dates?

Qqwy:
> I like where this is going. Is there any way we can use Date.add/2 (and eventually DateTime.add/3 and friends) here instead of converting between days and dates?

I think this should be possible, at least for most of the cases. :+1:

whatyouhide:
I’m closing this because it's been years 😄 We have date generators so far at least.

PR #167, opened 2022-01-24
Fix readability issue raised by Credo

kianmeng:


whatyouhide:
Hey @kianmeng, thanks for the PR. We don't use Credo in this repository. None of the changes made in this PR are "readability issues" except for the definition of "issue" used by Credo 🙃 

Making this change without setting up Credo with a reproducible configuration, CI, and so on can be pointless because no other contributors (including me) will run Credo on the codebase. I hope all of this made sense 🙃 

Thanks anyways! 💟

kianmeng:
@whatyouhide I see. Thanks for letting me know and what you said makes sense. I was wondering about that when I first checked with Credo.

By the way, okay to submit PR on missing test cases while I study how to use this library? :smile:

whatyouhide:
@kianmeng yes absolutely, missing test cases are always more than welcome! 💟

PR #168, opened 2022-01-27
Add property tests for the standard library's Enum.slide/3

s3cur3:
These are the property tests I wrote while developing `Enum.slide/3`. Following @whatyouhide's [invitation](https://mobile.twitter.com/whatyouhide/status/1484547725429555200), I thought it'd be nice to have these run daily against the latest version of the standard library. 😄

I don't think the stdlib tests run against every PR to `stream_data`, but [they've passed on my fork](https://github.com/s3cur3/stream_data/runs/4971869806?check_suite_focus=true).

### DIFF
--- a/test/stdlib/enum_test.exs
+++ b/test/stdlib/enum_test.exs
@@ -1,0 +1,77 @@
+defmodule StreamData.EnumTest do
+  use ExUnit.Case, async: true
+  use ExUnitProperties
+
+  @moduletag :stdlib
+
+  describe "Enum.slide/3" do
+    property "handles negative indices" do
+      check all(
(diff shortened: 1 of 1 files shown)


s3cur3:
Oh, uh… hehe. This will of course fail prior to 1.13. 🤔

whatyouhide:
@s3cur3 you can add conditional testing for that, something along these lines:

```elixir
# TODO: Remove when we depend on Elixir 1.13+.
if function_exported?(Enum, :slide, 3) do
  # What you have now
  property ...
end
```

s3cur3:
Aha, thanks for the tip! 😄 Fix pushed. 👍

whatyouhide:
Awesome, thanks @s3cur3! 💟

PR #170, opened 2022-05-29
Fix nonempty_improper_list_of/2 generator

### LINKED ISSUE #169: nonempty_improper_list_of generator (and by extension maybe_improper_list_generator) are incorrect
The generator

```elixir
StreamData.nonempty_improper_list_of(StreamData.integer(), StreamData.boolean())
```
has quite a high chance to generate `[true]` or `[false]`, but it should never do so.
`[true]` and `[false]` are not improper lists, but proper lists (syntactic sugar for `[true | []]` resp `[false | []]`).

I think the case over here:
https://github.com/whatyouhide/stream_data/blob/e8fd3355b38317a235929745663df32904bcf4d7/lib/stream_data.ex#L1100

is incorrect.
StreamData should not generate empty lists (and then this case will never happen).

Qqwy:
Ensures `[terminator]` will not be generated anymore by `nonempty_improper_list_of/2`

Fixes #169

### DIFF
--- a/test/stream_data_test.exs
+++ b/test/stream_data_test.exs
@@ -446,6 +446,7 @@ defmodule StreamDataTest do
   property "nonempty_improper_list_of/2" do
     check all list <- nonempty_improper_list_of(integer(), constant("")) do
       assert list != []
+      assert list != [""]
       each_improper_list(list, &assert(is_integer(&1)), &assert(&1 == ""))
     end
   end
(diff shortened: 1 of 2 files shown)


test/stream_data_test.exs
```diff
@@ -446,12 +446,14 @@ defmodule StreamDataTest do
   property "nonempty_improper_list_of/2" do
     check all list <- nonempty_improper_list_of(integer(), constant("")) do
       assert list != []
+      assert list != [""]
```
whatyouhide, line 449: I think the test you want here to be more precise is

```suggestion
      refute match?([_], list)
```

Qqwy:
The CI run is failing, but not on a test that has to do with this PR it seems.

PR #171, opened 2022-06-21
Support mixed optional/required key maps

zachdaniel:
By taking an optional second argument to `optional_map` we can support a mixed map of optional and required keys easily.


zachdaniel:
I may have just missed something but I couldn't find a helper that does this, and when I first started using it it was difficult to find an elegant way to do it myself. This seemed like a pretty common thing people would want to do, so I thought I'd PR.

whatyouhide:
@zachdaniel an "elegant" way to do this could be to generate two maps, the one with all the required keys, and then one with all the optional keys, and then `Map.merge/2` them. For example:

```elixir
{fixed_map(%{...}), optional_map(%{...}) |> map(fn {req, opt} → Map.merge(req, opt) end)
```

Would this work?

zachdaniel:
That works (and I arrived at it eventually) but it seems like a common enough need to potentially be worth being part of the API.

zachdaniel:
No worries if you disagree though :)

zachdaniel:
Well, I'm certainly glad I did, because the testing I did in my own app was clearly incorrect. The original implementation was incorrect. I can't see how the test failure that the build just got would be related to my code though...

test/stream_data_test.exs
```diff
@@ -527,6 +527,28 @@ defmodule StreamDataTest do
...
+        assert map_size(map) in [1, 2]
+
+        assert map
+               |> Map.keys()
+               |> MapSet.new()
+               |> MapSet.subset?(MapSet.new([:integer, :binary]))
+
+        if Map.has_key?(map, :integer) do
+          assert is_integer(Map.fetch!(map, :integer))
+        end
```
whatyouhide, line 545: This is  tricky test. It will pass if the generated map never has the `:integer` key.

I think a better test here would be to take a significant amount of values from the generator (with `Enum.take/2` if you want), like 100 values, and check that at least one generated map has the `:integer` key. It's statistically almost sure that it will 🙃

zachdaniel: 👍🏻 makes perfect sense, will do!

zachdaniel: Done :)

whatyouhide:
Thank you @zachdaniel! 💟

PR #174, opened 2022-06-23
add `StreamData.repeatedly/1`

### LINKED ISSUE #172: Simple generator for random/constant/unshrinkable things
I can't seem to find a way to make a simple generator for something like `&Ecto.UUID.generate/0`.  Per #38, `bind/2` works I think because the inner generator is recreated each time, so my best guess on implementing that generator is currently

```elixir
# Just picking any generator here
StreamData.integer()
|> StreamData.bind(fn _i -> StreamData.constant(Ecto.UUID.generate()) end)
|> StreamData.unshrinkable()
```

It seems strange to have to build off of another generator for no good reason there.

```elixir
StreamData.constant(Ecto.UUID.generate()) |> Enum.take(10)
```
This doesn't work, as naturally the list simply repeats the generated uuid.

zachdaniel:
`StreamData.repeatedly/1` is an unshrinkable generator that calls the
provided zero argument function to generate values

I'm honestly not entirely sure how `LazyTree` works, although I have a basic idea, but Im hoping I have the right idea here.

closes #172

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -715,7 +715,7 @@ defmodule StreamData do
   By nature, this generator is not shrinkable.
   """
   @spec repeatedly((arg :: any -> returns)) :: t(returns) when returns: term()
-  def repeatedly(fun) do
+  def repeatedly(fun) when is_function(fun, 0) do
     new(fn _seed, _size ->
       %LazyTree{root: fun.()}
     end)
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -694,6 +694,34 @@ defmodule StreamData do
   end
 
   @doc """
+  Calls the provided zero argument function to generate values.
+
+  ## Examples
+
+  Generating a UUID
+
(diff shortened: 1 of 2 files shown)


lib/stream_data.ex
```diff
@@ -693,6 +693,34 @@ defmodule StreamData do
...
+      integer = StreamData.repeatedly(&System.unique_integer([:positive, :monotonic]))
+      Enum.take(integer, 3)
+      #=> [1, 2, 3]
+
+  ## Shrinking
+
+  By nature, this generator is not shrinkable.
+  """
+  @spec repeatedly((arg :: any -> returns)) :: t(returns) when returns: term()
+  def repeatedly(fun) do
```
whatyouhide, line 718:
```suggestion
  def repeatedly(fun) when is_function(fun, 0) do
```

whatyouhide (approved):

whatyouhide:
Thank you @zachdaniel! 💟

PR #175, opened 2022-08-03
Add some properties for the Enum module

sabiwara:
This PR is adding some property-based tests for some of the `Enum` functions that had some bugs.

Since the implementation is often duplicated for lists, ranges and reduce-based enums like stream, a nice thing to test is that a function works the same on:
- a range
- a list made from this range
- a stream made from this range

This can catch inconsistencies and bugs, and we can easily add more coverage.


whatyouhide:
Fantastic @sabiwara ❤️

PR #176, opened 2022-08-04
Add property tests for Kernel.binary_slice/*

sabiwara:
This the test which caught https://github.com/elixir-lang/elixir/pull/12045


whatyouhide:
Awesome! 💟

PR #179, opened 2022-09-21
Fix dialyzer warnings inside project

### LINKED ISSUE #130: Generators make Dialyzer complain that a function has no local return

dmorneau:
Hi! This fixes warnings when running dialyzer against the project. Most are self-explanatory.

One that was a bit unintuitive is `ExUnitProperties.pick/1`:

```
lib/ex_unit_properties.ex:635: Invalid type specification for function 'Elixir.ExUnitProperties':pick/1.
 The success typing is 'Elixir.ExUnitProperties':pick(atom() | tuple() | #{'__struct__':='Elixir.StreamData', 'generator':=fun((_,_) -> any()), _=>_}) -> any()
 But the spec is 'Elixir.ExUnitProperties':pick('Elixir.StreamData':t(a)) -> a when a :: term()
 They do not overlap in the 1st argument
```

(this is on Erlang master, with improved errors: https://github.com/erlang/otp/pull/6271)

Dialyzer complains that pick's argument must be an instance of `atom() | tuple() | blabla...` for the function to succeed, and yet the typespec says the first argument is a `StreamData.t(a)`, which has nothing to do with the above in Dialyzer's mind (it's an opaque type, so it could potentially be anything - we can't make assumptions).

Why does Dialyzer think pick's argument has to be that specific type to succeed, and not a `StreamData.t(a)`? It's because pick calls `StreamData.__call__/3`, which lacks a typespec. Dialyzer infers the type of call's arguments by code analysis, giving the `atom() | ...` above. If we tell it via typespec that the argument of call is actually a `StreamData.t(a)`, then it figures out that pick will succeed if passed an argument of that type, and the warning goes away.

A similar problem happens when we call `list_of/1` or `atom/1`, which don't have typespecs, and pass the result to `bind_filter`, which expects the opaque type. I added missing typespecs and now get a clean pass on my app, so this might resolve #130. If there are more errors, I can take a look later.

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -540,6 +540,7 @@ defmodule ExUnitProperties do
     end
   end
 
+  @spec __raise__(term()) :: no_return()
   def __raise__(test_result) do
     %{
       original_failure: original_failure,
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -305,6 +305,10 @@ defmodule StreamData do
   raised. See the documentation for `filter/3` for suggestions on how to avoid
   such errors.
 
+  The function can accept one or two arguments. If a two-argument function is
+  passed, the second argument will be the number of tries left before raising
+  `StreamData.FilterTooNarrowError`.
+
   ## Examples
 
   Say we wanted to create a generator that generates two-element tuples where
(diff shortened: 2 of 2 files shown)


whatyouhide:
This is fantastic, awesome catches! 💟 I'll try to take some time to add Dialyzer to CI so that we avoid regressions. I'll release a new version as soon as we have that.

PR #180, opened 2022-10-14
Documentation update: Capture the generated user generator

chriseyre2000:
It took me some time to understand this example.
Adding a capture of the user generator makes it far clearer.

Thanks for having such as useful project!


whatyouhide:
Thank you for the documentation update and thanks for your patience, took me a while to get to this 🙈

PR #185, opened 2023-06-15
Add dialyxir to the mix (and CI)

paulo-ferraz-oliveira:
Hi.

Doing a couple of updates on libraries depending on this one, and wanna make sure I'm in the right place to continue the updates, since most of the issues I'm having are Dialyzer -based. Not from this library, though (at least not yet 😄), and I thought it wouldn't hurt to add `dialyxir` to it, for increased robustness.

Cheers.


.github/workflows/main.yml
```diff
@@ -52,6 +52,9 @@ jobs:
         if: ${{ matrix.lint }}
         run: mix format --check-formatted
 
+      - name: Run Dialyzer
+        run: mix dialyzer
```
whatyouhide, line 56: I'm ok with this, but we need to cache the PLTs built by Dialyzer. Something like [this](https://github.com/whatyouhide/redix/blob/3ba88b96e08ceab95e0839fc48f58adf488709dd/.github/workflows/main.yml#L56-L68) 🙃

mix.exs
```diff
@@ -47,8 +52,9 @@ defmodule StreamData.Mixfile do
 
   defp deps() do
     [
-      {:ex_doc, "~> 0.19", only: :dev},
-      {:excoveralls, "~> 0.16.0", only: :test}
+      {:ex_doc, "~> 0.29", only: :dev},
+      {:excoveralls, "~> 0.16.0", only: :test},
+      {:dialyxir, "~> 1.3", only: [:dev], runtime: false}
```
whatyouhide, line 57:
```suggestion
      {:dialyxir, "~> 1.3", only: [:dev, :test], runtime: false}
```

.github/workflows/main.yml
paulo-ferraz-oliveira: Surely :) I just didn't wanna go full-in until I understood if there was a chance to accept the change. Will take care. Thanks.

paulo-ferraz-oliveira:
@whatyouhide, one of the reasons that lead me to this was that `makeup`'s [lexer_stream_data_generators.ex](https://github.com/elixir-makeup/makeup/blob/master/test/helpers/lexer_stream_data_generators.ex#L22) is consuming `StreamData.one_of` with a non-opaque type, which issues a Dialyzer warning there. I was wondering if it would be better to (I can do it in a separate pull request, no prob.):
1. either have a function to make an input opaque (which seems kinda odd), something like
```elixir
@spec make_opaque([any()]) :: [t(a)] when a: term()
def make_opaque(t), do: t
```
2. or make the `@opaque t(a)` "public" (which is most likely undesired)
3. or have another type of function that helps build the opaque struct

.github/workflows/main.yml
paulo-ferraz-oliveira: c342a98

paulo-ferraz-oliveira:
I also pushed bdb9c03 because `dialyxir`'s default analysis scope is too broad...

paulo-ferraz-oliveira:
I've reversed the `mix.lock` changes since they were uninteresting in this PR's scope. Will wait for a `makeup_erlang` release to have it updated (in a different pull request).

paulo-ferraz-oliveira:
> Will wait for a makeup_erlang release to have it updated (in a different pull request).

Since this is still under review (and the changes aren't major) I pushed the [expected `ex_doc` update](https://github.com/whatyouhide/stream_data/pull/185/commits/3e60ebf39f499c3b34430fc62f6746e907e44dd4).

paulo-ferraz-oliveira:
Not sure the failing status is due to my changes...

paulo-ferraz-oliveira:
@whatyouhide, think it's possible to re-execute CI and then potentially merge? Or is anything missing? Thanks.

paulo-ferraz-oliveira:
Tests fail for Elixir 1.5.3 / Erlang 20.3. Want me to look at that, or could this pair be removed from CI?

whatyouhide:
@paulo-ferraz-oliveira I bumped the Elixir requirement, you can rebase this PR 🙃

paulo-ferraz-oliveira:
Rebased and force-pushed. Will tweak CI files...

paulo-ferraz-oliveira:
This should be good to go. Thanks for your help.

whatyouhide:
Thanks @paulo-ferraz-oliveira! 💟

PR #187, opened 2023-07-09
Add :utf8 option to string/1 and add chardata/0

sabiwara:
As discussed, existing [options](https://hexdocs.pm/stream_data/StreamData.html#string/2) can only generate subsets of valid strings (e.g. `printable`), this PR adds a `:utf8` option generating all of them.

It also adds a `chardata/0` generator, similar to `iodata/0` and relying on the `utf8` generator above.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -1818,6 +1818,10 @@ defmodule StreamData do
   end
 
   @ascii_chars ?\s..?~
+
+  # "UTF-8 prohibits encoding character numbers between U+D800 and U+DFFF"
+  @utf8_chars [0..0xD7FF, 0xE000..0x10FFFF]
+
   @alphanumeric_chars [?a..?z, ?A..?Z, ?0..?9]
   @printable_chars [
     ?\n,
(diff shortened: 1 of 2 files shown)


lib/stream_data.ex
```diff
@@ -1818,6 +1818,10 @@ defmodule StreamData do
   end
 
   @ascii_chars ?\s..?~
+
+  # "UTF-8 prohibits encoding character numbers between U+D800 and U+DFFF"
```
sabiwara, line 1822: https://www.rfc-editor.org/rfc/rfc3629

lib/stream_data.ex
```diff
@@ -2042,6 +2047,40 @@ defmodule StreamData do
...
+      Enum.take(StreamData.chardata(), 3)
+      #=> ["", [""], [12174]]
+
+  ## Shrinking
+
+  Shrinks towards less nested chardata and ultimately towards smaller binaries.
+  """
+  @spec chardata() :: t(IO.chardata())
+  def chardata() do
+    codepoint = @utf8_chars |> Enum.map(&{Enum.count(&1), integer(&1)}) |> frequency()
```
sabiwara, line 2066: I guess there could be a case for a `codepoint/0` generator, let me know your thoughts

whatyouhide: Yep, we can add it in another PR if that's ok with you!

whatyouhide:
Fantastic, thanks @sabiwara! 💟

lib/stream_data.ex
sabiwara: Sure, I'm on it!

sabiwara: Reconsidering it, should it rather be `codepoint/1` and accept `:utf8 | :printable | :ascii | :alphanumeric`? 🤔

whatyouhide: Yes, that sounds good 👍

PR #188, opened 2023-07-09
Extract codepoint/1 generator

sabiwara:
Note: the second [commit](https://github.com/whatyouhide/stream_data/commit/2835e8c3e599e7438a3483d8beb18445178d8356) fixes the error message that I forgot in the previous PR https://github.com/whatyouhide/stream_data/pull/187.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -1840,7 +1840,7 @@ defmodule StreamData do
   ]
 
   @doc """
-  Generates an integer corresponding to a valid UTF8 codepoint of the given kind.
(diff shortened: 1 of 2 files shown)


lib/stream_data.ex
```diff
@@ -1839,6 +1839,54 @@ defmodule StreamData do
     0x10000..0x10FFFF
   ]
 
+  @doc """
+  Generates an integer corresponding to a valid UTF8 codepoint of the given kind.
```
whatyouhide, line 1843:
```suggestion
  Generates an integer corresponding to a valid UTF-8 codepoint of the given kind.
```

PR #192, opened 2024-02-12
Fix deprecation warning on register_test/4

sabiwara:
On Elixir's `main` branch, `ExUnit.Properties` emits deprecation warnings on [`ExUnit.Case.register_test/4`](https://hexdocs.pm/ex_unit/1.16.1/ExUnit.Case.html#register_test/4):

Screenshot 2024-02-12 at 13 33 50

Since `StreamData` supports Elixir `~> 1.11` and `register_test/6` is `since: "1.10.0"`, we can fix it by replacing`register_test/4` by its [implementation](https://github.com/elixir-lang/elixir/blob/v1.16.1/lib/ex_unit/lib/ex_unit/case.ex#L635).

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -220,7 +220,8 @@ defmodule ExUnitProperties do
     contents = Macro.escape(contents, unquote: true)
 
     quote bind_quoted: [context: context, contents: contents, message: message] do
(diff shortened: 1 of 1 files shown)


whatyouhide:
Oh sweet good catch @sabiwara! Thanks 💟

PR #194, opened 2024-05-19
Updates version in README to 1.0

vrcca:
It was still showing 0.6


whatyouhide (commented):
I don't like stressing the Hex API with unnecessary calls. The link to Hex was like that on purpose, so that it's just a static link. The version is not necessary to have in the README.

Can you revert those badges changes and just leave the other update? Thanks!

whatyouhide (approved):

whatyouhide:
Thank you 💟

PR #195, opened 2024-05-24
Add an explicit step to a range match

alco:
Without this, Elixir 1.17 emits a warning for this line.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -963,7 +963,7 @@ defmodule StreamData do
         {:ok, length} when is_integer(length) and length >= 0 ->
           {length, length}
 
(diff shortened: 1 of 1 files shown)


whatyouhide:
CI is failing, mind taking a look? We might have to do this conditionally based on the Elixir version, I’m not sure. I'll be without a laptop til Monday so I can't check til then.

alco:
@whatyouhide I have rewritten the pattern matching to use the `Range` struct. If that looks to weird, then never mind. It's not a critical warning to address of compatibility with Elixir v1.11 is a must.

whatyouhide:
Actually, yeah that's a good point. I just dropped support for Elixir 1.11 and fixed this in the latest commit. Thanks @alco 💟

PR #199, opened 2024-09-29
Fix warning about range step

maennchen:
Fixes the following warning:

```
warning: Range.new/2 has a default step of -1, please call Range.new/3 explicitly passing the step of -1 instead
  (stream_data 1.1.1) lib/stream_data.ex:1723: anonymous fn/3 in StreamData.power_of_two_with_zero/1
```


whatyouhide:
Thanks @maennchen! 💟

PR #200, opened 2024-10-17
Remove check for __STACKTRACE__ support

davydog187:
StreamData relies on Elixir 1.12 now, so this code is no longer needed


whatyouhide:
💟

PR #201, opened 2024-10-17
Remove deprecated :exs64

davydog187:
The minimum version of Elixir for StreamData is 1.12, which requires OTP 22+


whatyouhide:
💟

PR #202, opened 2024-10-17
Use ranges with step in tests

davydog187:
StreamData depends on Elixir 1.12+ now, so we can clean these up


whatyouhide:
Sweet, thanks Dave 💟

PR #204, opened 2024-10-31
Fix block code indent in doc

sabiwara:
Before: https://hexdocs.pm/stream_data/StreamData.html#non_negative_integer/0

After:
Screenshot 2024-10-31 at 11 05 00


whatyouhide:
Thanks @sabiwara 💟

PR #206, opened 2025-01-26
doc: indent example to fix display

ahamez:
Fix the following incorrect display: CleanShot 2025-01-26 at 16 57 01@2x


whatyouhide:
Thanks @ahamez  💟

Issue #213, opened 2025-05-05
Option to expand generated value inspect


andrewtimberlake:
I’m using StreamData and ExUnitProperties to generate a very large data structure. When it fails, it would be useful to be able to have it inspect the value with `inspect(value, limit: :infinity)` so I can find the problem even though it might be located deep in the structure.

https://github.com/whatyouhide/stream_data/blob/48a3e602b61053c5565721b251d199ecc264aeac/lib/ex_unit_properties.ex#L640

whatyouhide:
@andrewtimberlake yes please 💟

andrewtimberlake:
Merged in #214

PR #214, opened 2025-05-06
Add ability to customise the inspect options used

andrewtimberlake:
Adds `:inspect_opts` config as per #213

### DIFF
--- a/lib/ex_unit_properties.ex
+++ b/lib/ex_unit_properties.ex
@@ -133,6 +133,20 @@ defmodule ExUnitProperties do
   examples. Fred's website uses an Erlang property-based testing tool called
   [PropEr](https://github.com/manopapad/proper) but many of the things he talks
   about apply to `ExUnitProperties` as well.
+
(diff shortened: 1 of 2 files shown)


lib/ex_unit_properties.ex
```diff
@@ -637,7 +651,7 @@ defmodule ExUnitProperties do
     Enum.map_join(values, "\n\n", fn {gen_string, value} ->
       String.trim_trailing("""
       * Clause:    #{gen_string}
-        Generated: #{inspect(value)}
+        Generated: #{inspect(value, Application.get_env(:stream_data, :inspect_opts, []))}
```
whatyouhide, line 654: Let's add the default (`[]`) to `application/0` in `mix.exs` and call `Application.fetch_env!/2` here instead.

andrewtimberlake: Done. I learned something new today ;-)

whatyouhide (approved):

PR #218, opened 2025-08-24
Date generators

Munksgaard:
This is a reimplementation of the date generators by @Qqwy in #161. As such, it is a (partial) implementation for #129.

The tests and documentation are entirely copy-pasted from his PR, but the actual implementation is entirely new. As noted by @whatyouhide in a comment to the original PR, I've opted to use the new `Date.add` function for most of the cases.

I think the resulting implementation is a bit easier to follow. I have not added Time, DateTime or NaiveDateTime, though I think we could probably do something for those that follow this skeleton.

### DIFF
--- a/lib/stream_data.ex
+++ b/lib/stream_data.ex
@@ -2095,6 +2095,90 @@ defmodule StreamData do
   end
 
   @doc """
+  Generates dates according to the given `options` or `date_range`.
+
+  ## Options
+
+    * `:origin` - (`t:Date.t/0`) if present, generated values will shrink towards this date.
+      Cannot be combined with `:min` or `:max`.
+
+    * `:min` - (`t:Date.t/0`) if present, only dates *after* this date will be
+      generated. Values will shrink towards this date.
+
+    * `:max` - (`t:Date.t/0`) if present, only dates *before* this date will be
+      generated. Values will shrink towards this date.
+
+  If both `:min` and `:max` are provided, dates between the two mentioned dates will be generated.
+  Values will shrink towards `:min`.
+
+  If no options are provided, `:origin` will default to the current date (that is,
+  `Date.utc_today/0`).
+
+  ## `t:Date.Range.t/0`
+
+  Alternatively a `t:Date.Range.t/0` can be given. This will generate dates in the given range,
+  and with the supplied `date_range.step`.
+
+  Values will shrink towards `date_range.first`.
+
+  ## Calendar Support
+
+  This generator works with `m:Calendar.ISO` and any other calendar
+  which implements the callbacks
+  `c:Calendar.naive_datetime_to_iso_days/7` and `c:Calendar.naive_datetime_from_iso_days/2`.
+  """
+  @spec date(Date.Range.t() | keyword()) :: t(Date.t())
+  def date(options_or_date_range \\ [origin: Date.utc_today()])
+
+  def date(date_range = %Date.Range{}) do
+    Date.to_gregorian_days(date_range.first)..Date.to_gregorian_days(date_range.last)//date_range.step
+    |> StreamData.integer()
+    |> StreamData.map(&Date.from_gregorian_days(&1))
+  end
+
+  def date(options) when is_list(options) do
+    min = Keyword.get(options, :min, nil)
+    max = Keyword.get(options, :max, nil)
+    origin = Keyword.get(options, :origin)
+
+    case {min, max, origin} do
+      {nil, nil, nil} ->
+        StreamData.integer() |> StreamData.map(&Date.add(Date.utc_today(), &1))
+
+      {nil, nil, origin = %Date{}} ->
+        StreamData.integer() |> StreamData.map(&Date.add(origin, &1))
+
+      {nil, max = %Date{}, nil} ->
+        StreamData.positive_integer() |> StreamData.map(&Date.add(max, -&1))
+
+      {min = %Date{}, nil, nil} ->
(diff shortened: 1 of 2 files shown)


lib/stream_data.ex
```diff
@@ -2094,6 +2094,85 @@ defmodule StreamData do
...
 
+  @doc """
+  Generates arbitrary (past and future) dates.
+
+  Generated values shrink towards `Date.utc_today/0`.
+  """
+  @spec date() :: t(Date.t())
+  def date do
+    date([])
+  end
```
whatyouhide, line 2105: Can we remove this in favor of `[]` options in the other clause? That way for example we can say `:origin` *defaults* to `Date.utc_today/0`.

lib/stream_data.ex
```diff
@@ -2094,6 +2094,85 @@ defmodule StreamData do
...
+
+  @doc """
+  Generates dates according to the given `options` or `date_range`.
+
+  ## Options
+
+  * `:origin` - (`Date`) if present, generated values will shrink towards this date. Cannot be combined with `:min` or `:max`.
```
whatyouhide, line 2112: Let's replace types with links, here it would be:


```suggestion
  * `:origin` - (`t:Date.t/0`) if present, generated values will shrink towards this date. Cannot be combined with `:min` or `:max`.
```

lib/stream_data.ex
```diff
@@ -2094,6 +2094,85 @@ defmodule StreamData do
...
+  If no options are provided, will work just like `date/0`.
+
+  ## `Date.Range`
+
+  Alternatively a `Date.Range` can be given. This will generate dates in the given range,
+  and with the supplied `date_range.step`.
+
+  Values will shrink towards `date_range.first`.
+
+  ## Calendar support
```
whatyouhide, line 2130:
```suggestion
  ## Calendar Support
```

lib/stream_data.ex
```diff
@@ -2094,6 +2094,85 @@ defmodule StreamData do
...
+      {min = %Date{}, max = %Date{}, nil} ->
+        if Date.before?(max, min) do
+          raise ArgumentError, "max must not be before min"
+        end
+
+        diff = Date.diff(max, min)
+        StreamData.integer(0..diff) |> StreamData.map(&Date.add(min, &1))
+
+      _ ->
+        raise ArgumentError, "invalid arguments"
```
whatyouhide, line 2172: Let's explain why

lib/stream_data.ex
```diff
@@ -2094,6 +2094,85 @@ defmodule StreamData do
...
+
+      {nil, max = %Date{}, nil} ->
+        StreamData.positive_integer() |> StreamData.map(&Date.add(max, -&1))
+
+      {min = %Date{}, nil, nil} ->
+        StreamData.positive_integer() |> StreamData.map(&Date.add(min, &1))
+
+      {min = %Date{}, max = %Date{}, nil} ->
+        if Date.before?(max, min) do
+          raise ArgumentError, "max must not be before min"
```
whatyouhide, line 2165: Let's show the given options.


```suggestion
          raise ArgumentError, """
          expected :max to be after or equal to :min, got:
          
            * min: #{inspect(min)}
            * max: #{inspect(max)}
          """
```

lib/stream_data.ex
Munksgaard: Got it.

lib/stream_data.ex
Munksgaard: Done.

lib/stream_data.ex
Munksgaard: Makes sense.

lib/stream_data.ex
```diff
@@ -2094,6 +2094,80 @@ defmodule StreamData do
...
+  @doc """
+  Generates dates according to the given `options` or `date_range`.
+
+  ## Options
+
+  * `:origin` - (`t:Date.t/0`) if present, generated values will shrink towards this date. Cannot be combined with `:min` or `:max`.
+
+  * `:min` - (`t:Date.t/0`) if present, only dates _after_ this date will be generated. Values will shrink towards this date.
+
+  * `:max` - (`t:Date.t/0`) if present, only dates _before_ this date will be generated. Values will shrink towards this date.
```
whatyouhide, line 2106:
```suggestion
    * `:origin` - (`t:Date.t/0`) if present, generated values will shrink towards this date.
      Cannot be combined with `:min` or `:max`.

    * `:min` - (`t:Date.t/0`) if present, only dates *after* this date will be
      generated. Values will shrink towards this date.

    * `:max` - (`t:Date.t/0`) if present, only dates *before* this date will be
      generated. Values will shrink towards this date.
```

lib/stream_data.ex
```diff
@@ -2094,6 +2094,80 @@ defmodule StreamData do
...
+
+  If both `:min` and `:max` are provided, dates between the two mentioned dates will be generated.
+  Values will shrink towards `:min`.
+
+  If no options are provided, `origin` will default to the current date (by calling `&Date.utc_today/0`).
```
whatyouhide, line 2111:
```suggestion
  If no options are provided, `:origin` will default to the current date (that is,
  `Date.utc_today/0`).
```

lib/stream_data.ex
whatyouhide: I meant in the error message 🙃

Munksgaard: I thought for sure I had done that. Sorry for the inconvenience, it should be updated now.

whatyouhide:
Thank you @Munksgaard 🫶

PR #219, opened 2025-08-25
Apply filter to float_with_bounds

### LINKED ISSUE #203: `float()` will return `0.0` as a valid value if `min` is a very small number and `max` is approaching 0 from the neg. side
Testing some edge cases I came across an unexpected behaviour of [float](https://github.com/whatyouhide/stream_data/blob/v1.1.2/lib/stream_data.ex#L1671)

`StreamData.float()` will include `0.0` when:
- negative `min` and `max`
- `min` approaching 0.0 (but not touching it)
- `max` bing a very big negative number.

```elixir
iex()> StreamData.float(min: -9.9e25, max: -0.1) |> Enum.take(10)
[0.0, -9.9e25, 0.0, -8.6625e25, -2.4749999999999994e25, -6.0328125e25, -9.9e25,
 -1.2374999999999997e25, -1.3921875e25, 0.0]
```
This should not contain any `0.0`

On the other side. These variations work as expected
```elixir
iex(0)> StreamData.float(min: -9, max: -0.1) |> Enum.take(10)
[-0.09999999999999964, -9.0, -0.09999999999999964, -0.09999999999999964, -9.0,
 -9.0, -6.21875, -0.09999999999999964, -7.8875, -3.715625]

iex()> StreamData.float(max: -0.1) |> Enum.take(10)
[-0.1, -5.1, -0.1, -0.6625, -1.0375, -0.85, -1.084375, -61.4203125,
 -397.414453125, -21.90859375]

iex()> StreamData.float(min: 0.1, max: 9.9e25) |> Enum.take(10)
[4.95e25, 9.9e25, 6.1874999999999994e25, 9.9e25, 0.1, 5.878125e25, 9.9e25,
 3.673828125e25, 2.475e25, 5.64609375e25]

iex()> StreamData.float(min: 0.1) |> Enum.take(10)
[1.1, 4.1, 0.85, 2.0375, 10.81875, 5.428125, 0.7015625, 0.271875, 384.887109375,
 0.7806640625]
```

Munksgaard:
Computing floats when given bounds such as min -9.9e15 and max -1 will result in the computation `-1.0 - -9.9e15` yielding 9900000000000000.0 instead of 9899999999999999.0, which makes the generator return 0. This is a classic floating point loss of significance error.

Fixes #203


whatyouhide:
This doesn't affect "normal" use cases where the range is much smaller, right? (Given the issue only appears with large numbers)

Munksgaard:
I've updated the PR to make it more likely that the special cases are hit in tests. Hopefully that should help fix the coveralls regression.

whatyouhide:
Superseded by #225. Thanks @Munksgaard! 🫶

PR #221, opened 2025-08-28
Add nullable/2 generator

### LINKED ISSUE #220: Feature request: Add `nilable` or `maybe` function
This is a minor thing, but I find myself writing `StreamData.one_of([nil, some_other_generator])` quite a lot. Would there be interest in a PR that adds a `StreamData.maybe/1` generator that makes another generator nilable?

Of course, there's a discussion about what the right distribution of values should be here. The simple `one_of` from above makes it so that `nil` is returned 50% of the time, which may or may not be the desired outcome. Feedback welcome!

Munksgaard:
As discussed in #220, this PR implements a `nullable/2` generator. It returns either the result of the generator given as the first argument or `nil`. The default ratio between the two possibilities is 0.5, but can be adjusted using the `:ratio` option.

Fixes #220


whatyouhide:
> This is probably a more controversial suggestion, which is why I haven't added it in this PR, but I'm wondering if there would be interest in it?

I think the API surface is already large and that doesn't seem like a very useful addition. Also same as:

```elixir
[some, other, gen]
|> one_of()
|> nullable()
```

whatyouhide:
> I don't think that is the same as what I was suggesting. In your code, nil would have 50% probability of being returned. In my suggested code it would have 25% probability of being returned.

Solved by `|> nullable(ratio: 0.25)`? or `|> nullable(ratio: length(generators) / (length(generators) + 1))` or something.

> I don't think specifying a ratio of 1.0 make sense.

Then `0.0` doesn't make sense either, and in any case, allowing both makes it easier to have that ratio come from other parts of the code if necessary. So yes, let's relax this.

Munksgaard:
> Solved by `|> nullable(ratio: 0.25)`? or `|> nullable(ratio: length(generators) / (length(generators) + 1))` or something.

Correct.

> Then 0.0 doesn't make sense either, and in any case, allowing both makes it easier to have that ratio come from other parts of the code if necessary. So yes, let's relax this.

0.0 also wasn't allowed, but I've pushed a new commit that relaxes the constraints.

whatyouhide:
Thanks @Munksgaard 💟

